PR #11, opened 2021-11-25
[WIP] Implementing batching

andrenatal:
Remarkable improvement of performance in terms of wps (am seeing north of 500) and user experience, but it breaks the heuristics currently being used for tag manipulation.

So it needs https://github.com/browsermt/bergamot-translator/pull/253 in order to work flawless.

cc @abhi-agg @eu9ene


abhi-agg:
This is great. 500 wps using wormhole enabled artifacts?

andrenatal:
> This is great. 500 wps using wormhole enabled artifacts?

Those artifacts we couldn't confirm if it was using it not.

andrenatal:
Could notice 1000wps in some cases

eu9ene:
> Could notice +1000wps in some cases

cool stuff! yes, it makes a lot of sense to use batching for inference

extension/controller/translation/Translation.js
```diff
@@ -9,6 +9,10 @@
 class Translation {
     constructor (mediator){
         this.translationsMessagesCounter = 0;
+        this.TRANSLATION_INTERVAL = 100; // ms
+        this.MAX_TRANSLATION_MSGS = 500; // max translations to process per batch
```
eu9ene, line 13: how did you pick this number? maybe we should play with it and see what's optimal, it might also depend on the user's hardware

andrenatal: just randomly experimenting, if we could find an equation here based on the platform factors would be the ideal if we have the tools available

PR #38, opened 2022-01-20
Telemetry

eu9ene:
Basic telemetry integration:
- v1 schema (everything is in one ping which is not too bad because it reduces the number of pings, simplifies sending and doesn't require expensive joins on analysis)
- simple and lightweight glean client, that imitates behavior of glean.js 0.15.0 ( the one that was integrated in the legacy extension)
- the most important in-page translation metrics are collected

Out of scope for this PR:
- #42 
- #43 
- #37 
- #45 
- edge cases when ping is not sent (browser event wasn't triggered) - to find


extension/mediator.js
```diff
@@ -27,7 +32,16 @@ class Mediator {
     }
 
     init() {
-        browser.runtime.sendMessage({ command: "monitorTabLoad" });
+        this.telemetry.loadSchema()
+            .then(() => {
+                browser.runtime.sendMessage({ command: "monitorTabLoad" })
```
eu9ene, line 37: sometimes mediator.start is called before telemetry schema is loaded on page refresh. I don't understand why this happens, probably old subscription to "responseMonitorTabLoad" in background script survives the refresh.

extension/controller/backgroundScript.js
```diff
@@ -7,6 +7,10 @@
  * this script does not have access to the page content, since it runs in the
  * extension's background process.
  */
+
+let cachedEnvInfo = null;
+console.debug("bg script loaded");
```
andrenatal, line 12: can we remove this?

extension/mediator.js
```diff
@@ -28,12 +33,19 @@ class Mediator {
...
+
+    // the page is closed or infobar is closed manually
+    closeSession() {
+        this.telemetry.submit("custom");
     }
 
     // main entrypoint to handle the extension's load
     start(tabId) {
-
+        console.debug("starting mediator");
```
andrenatal, line 46: remove this

extension/mediator.js
```diff
@@ -152,19 +199,24 @@ class Mediator {
...
+    // eslint-disable-next-line max-lines-per-function
     bgScriptsMessageListener(message) {
         switch (message.command) {
             case "responseMonitorTabLoad":
                 this.start(message.tabId);
                 break;
+            case "telemetryInfoLoaded":
+                this.telemetry.setBrowserEnv(message.env);
+                this.translationTelemetry.recordEnvironment(message.env);
+                this.translationTelemetry.recordVersions("0.5.0", "?", "?");
```
andrenatal, line 211: is this the addon version? maybe is better to read it from the manifest.

andrenatal (approved):
I didn't test it but it's looking good. For the next patches is better to have smaller PRs

andrenatal:
Just make sure to read the addon version from the manifest to automate that.

extension/mediator.js
eu9ene: I wanted to do this in #42

extension/mediator.js
eu9ene: removed

extension/controller/backgroundScript.js
eu9ene: removed

extension/mediator.js
eu9ene: done

andrenatal (approved):

PR #47, opened 2022-01-23
Do not display the translationbar to non-supported languagues

### LINKED ISSUE #34: Infobar is displayed for unsupported languages
Go to https://fr.wikipedia.org/wiki/Wikip%C3%A9dia:Accueil_principal

Language is detected correctly as French, but the infobar is displayed with an empty drop-down box.

It should not be displayed.

andrenatal:
fixes #34

### DIFF
--- a/extension/mediator.js
+++ b/extension/mediator.js
@@ -61,7 +61,7 @@ class Mediator {
         if (!this.languageDetection.navigatorLanguage.includes(this.languageDetection.pageLanguage.language)) {
 
             /*
-             *  todo: we need to keep track if the translationbar was already displayed
+             * we need to keep track if the translationbar was already displayed
              * or not, since during tests we found the browser may send the
              * onLoad event twice.
              */
--- a/extension/view/js/TranslationNotificationManager.js
+++ b/extension/view/js/TranslationNotificationManager.js
@@ -93,4 +93,8 @@ class TranslationNotificationManager {
         const message = { command: "displayStatistics", tabId: this.tabId };
         this.bgScriptListenerCallback(message);
     }
+
+    isPageLanguageSupported(){
+        return this.languageSet.has(this.detectedLanguage);
+    }
 }
--- a/extension/controller/experiments/TranslationBar/api.js
+++ b/extension/controller/experiments/TranslationBar/api.js
@@ -47,6 +47,15 @@
           translationbar: {
             show: function show(tabId, detectedLanguage, navigatorLanguage) {
               try {
+                translationNotificationManager = new TranslationNotificationManager(
+                  this,
+                  modelRegistry,
+                  detectedLanguage
+                );
+
+                if (!translationNotificationManager.isPageLanguageSupported()) {
+                  return
+                }
 
                 // disable the legacy translation infobar
                 Services.prefs.setBoolPref("browser.translation.ui.show", false);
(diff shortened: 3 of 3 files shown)


extension/view/js/TranslationNotificationManager.js
```diff
@@ -93,4 +93,8 @@ class TranslationNotificationManager {
         const message = { command: "displayStatistics", tabId: this.tabId };
         this.bgScriptListenerCallback(message);
     }
+
+    isPageLanguageSupported(){
+        return this.languageSet.has(this.detectedLanguage);
```
eu9ene, line 98: We should check that language pair page-lang->user-lang or page-lang->en and en->user-lang are supported. Models of opposite directions are completely decoupled from each other and we can support only one-way translation.

extension/mediator.js
```diff
@@ -61,7 +61,7 @@ class Mediator {
         if (!this.languageDetection.navigatorLanguage.includes(this.languageDetection.pageLanguage.language)) {
```
eu9ene, line 61: The code inside "if" statement shouldn't be called if language pair is not supported. It triggers calls to infobar and creates unnecessary Translation object. Also we'll need to record "unsupported" metrics from mediator. It means we need to read model registry from mediator and move the check here from TranslationNotificationManager.

eu9ene (changes requested):

extension/view/js/TranslationNotificationManager.js
andrenatal: After giving this a second thought there's some assumptions we are making about this issue for free that I'm not sure if they are right:

- We are putting to much trust that the cld module will return right language, and we already know how naive that module is and its flaws (like confusing es and pt very often)
- We are assuming that the portion of the page's content to be analyzed to always be representative of the majority of the page's language.

So I think this is more complicated than just preventing the infobar of displaying when the supposed page's language do not match what exists in the model registry. Before that we'll need to review the cld and the page's language analysis first and also decide if we want to actually prevent the infobar of displaying or not (maybe we want to always display it and only don't auto select the language we think it was detected? 

In a normal world that would be a UI + product decision, but since we know we don't have that, we'll need to make a decision ourselves.

extension/mediator.js
andrenatal: | Also we'll need to record "unsupported" metrics from mediator.

Could you please elaborate on what does this mean? That we are supposed to send a ping to every time a page is not supported by the extension (i.e. all the time the user browses to a new page?)

extension/view/js/TranslationNotificationManager.js
eu9ene: I think we should definitely work on improving cld accuracy but I don't think always displaying the infobar will be useful. Our language coverage is pretty limited atm and displaying just to show that we don't support the page language will be annoying for users. I would suggest assuming cld works correctly and relying on its output here, abstract from it and work on this issue separately.

extension/mediator.js
eu9ene: See this [todo](https://github.com/mozilla/firefox-translations/pull/38/files#diff-ce3e24cd36ecb87c17863aa23203ed33457bb09e419fd2dd192ba23576d24f50R88). We should always record "unsupported" metric. How the ping sending is triggered is decoupled from recording. In the current implementation, it happens on page closing. If it won't work for us, we can move this metric to a different ping and send it on timer in the future.

andrenatal: Ok, so this means sending a ping to every page a user browses to any page, no? At least one ping will always be sent  if the page's lang is either supported or unsupported.

eu9ene: Only when language mismatch happens. Yes, exactly one ping will be sent if the user language is different from the page language in the current implementation.

andrenatal:
> @andrenatal One question. Is there a test for this?



> @andrenatal One question. Is there a test for this?

@abhi-agg  please see https://github.com/mozilla/firefox-translations/issues/49 

Wanna take a stab at that?

eu9ene (approved):

Issue #51, opened 2022-01-26
non-HTML input is treated as HTML


jelmervdl:
It looks like [all input is marked as HTML for the translator](https://github.com/mozilla/firefox-translations/blob/c21b61fdd1dc3186bb478c7702d601dc20b7b336/extension/controller/translation/translationWorker.js#L496
), even though [nodes' text content is submitted](https://github.com/mozilla/firefox-translations/blob/c21b61fdd1dc3186bb478c7702d601dc20b7b336/extension/view/js/InPageTranslation.js#L135-L158). If a node contains something like `<p>Hello &lt; world</p>`, it would submit `Hello < world`. Which when parsed as HTML is invalid input and would cause an exception.

Even if HTML was being submitted, it would not be properly used (and cause an `abort()`) because the model doesn't produce alignment information. In [the model configuration yaml](https://github.com/mozilla/firefox-translations/blob/c21b61fdd1dc3186bb478c7702d601dc20b7b336/extension/controller/translation/translationWorker.js#L262-L275), the line `alignment: soft` is missing.

andrenatal:
@jelmervdl Maybe this is related to https://github.com/mozilla/firefox-translations/issues/35, where I'm seeing multiple exceptions in the engine but can't reproduce on different languages?

jelmervdl:
@andrenatal it could be very well related. There are paragraphs with literal `<` and `>` in them. Those paragraphs would cause the html parsing bit in bergamot-translator to throw (but because of how emscripten is configured… abort()?) since it tries to parse that as an open tag and fail.

kpu:
The extension will be called upon to translate `text/plain` and `text/HTML` documents, so the extension will need to toggle HTML on/off anyway, using this:

https://github.com/browsermt/bergamot-translator/blob/c0f311a8c067372057a6f301c42b40bbe30a9c1a/src/translator/response_options.h#L22
(The TODO comment is outdated and will be fixed.)

andrenatal:
@abhi-agg could you please take a look at this?

abhi-agg:
@andrenatal I am on it.

abhi-agg:
> The extension will be called upon to translate `text/plain` and `text/HTML` documents, so the extension will need to toggle HTML on/off anyway

@kpu By `documents`, do you literally mean the user will upload documents to translate or do you only mean web pages?

kpu:
[...]

PR #53, opened 2022-01-26
Add `alignment: soft;` to YAML

kpu:
This is required for HTML support to work.

### DIFF
--- a/extension/controller/translation/translationWorker.js
+++ b/extension/controller/translation/translationWorker.js
@@ -272,6 +272,7 @@ class TranslationHelper {
             quiet: true
             quiet-translation: true
             gemm-precision: int8shiftAlphaAll
+            alignment: soft
             `;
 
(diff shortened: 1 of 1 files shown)


andrenatal:
Which version of bergamot-translator is required to have for this flag? The one we are currently using is compatible with this? I'll let @abhi-agg review this

kpu:
It's been used in the demo since https://github.com/browsermt/bergamot-translator/blame/feb9c90429fe23423dc23c56e8d0ee19d85acec7/wasm/test_page/js/worker.js#L187

kpu:
The HTML feature has always depended on this.  So any quality of HTML you see in the extension until this is merged would indeed be terrible.  I think we should start enforcing it with an error.

abhi-agg (approved):

abhi-agg:
@andrenatal I verified that the artifacts that the extension currently uses already included this change. So, we are safe to merge it.

andrenatal:
@abhi-agg does this fixes #51 for free?

Issue #60, opened 2022-01-28
Translating non-German webpages to German doesn't work


abhi-agg:
Steps to reproduce:
1. Download Firefox Nightly (German version)
2. Start extension (latest main branch)
3. Navigate to page: https://andrenatal.com/translations-playground/?lang=pt
4. Infobar correctly displays the page language
5. Click on `Übersetzen`
6. The content is not translated

Replacing `pt` with `es`, `et`, `en`, `it`, `ru` and `cs` in the link https://andrenatal.com/translations-playground/?lang=pt gives same behavior with following error message in web console:

```
Error: We did not find an alpha in the model named: F0::Wemb_QuantMultA. bergamot-translator-worker.js:1217:12
Error: Aborted from auto marian::cpu::integer::fetchAlphaFromModelNodeOp::forwardOps()::(anonymous class)::operator()() const in /root/checkout/3rd_party/marian-dev/src/tensors/cpu/intgemm_interface.h:583 bergamot-translator-worker.js:1217:12
Callstacks not supported in WASM builds currently bergamot-translator-worker.js:1217:12
undefined bergamot-translator-worker.js:649:9
Translation error:  RuntimeError: abort(undefined). Build with -s ASSERTIONS=1 for more info. translationWorker.js:120:37
RuntimeError: abort(undefined). Build with -s ASSERTIONS=1 for more info. bergamot-translator-worker.js:653:14
```

Issue #58, opened 2022-01-27
`We did not find an alpha in the model named: F0::Wemb_QuantMultA.` when translating from `pt` to `de` on outbound translations. 


andrenatal:
I'm getting the crash below [2] in the model/engine when translating from `pt` to `de` using outbound translations (although this seems to be a generalized issues). It's not happening from `pt` to the other languages but it might be happening with different combinations. There's a screen record of the error here: 

STR:
1 - Download Nightly `pt-br`
2 - Load the extension from master
3 - Navigate to: `http://andrenatal.github.io/translations-playground`
4 - Choose `de`
5  - Click on `Traduzir`
6 - Click on Enable translation of form `Yes`
7 - Click on the textarea
8 - Type something in the textarea withing the outbound translations widget

[1] https://www.dropbox.com/s/7vw5vhykcnlnwjv/ptde_crash.mov?dl=1

[2] 
```
Using fallback gemm implementation bergamot-translator-worker.js:6245:17
Wasm Runtime initialized Successfully (preRun -> onRuntimeInitialized) in 0.011 secs translationWorker.js:67:29
Creating Translation Service with config: [object Object] translationWorker.js:230:25
Translation Service created successfully translationWorker.js:232:25
Constructing model 'dept' via pivoting: 'deen' and 'enpt' translationWorker.js:252:25
Total Download time for all files of 'enpt': 0.185 secs translationWorker.js:321:21
Constructing Aligned memory. Size: 17140836 bytes, Alignment: 256 translationWorker.js:481:21
Aligned memory construction done translationWorker.js:483:21
Aligned memory initialized translationWorker.js:486:21
Constructing Aligned memory. Size: 4472528 bytes, Alignment: 64 translationWorker.js:481:21
Aligned memory construction done translationWorker.js:483:21
Aligned memory initialized translationWorker.js:486:21
Constructing Aligned memory. Size: 812781 bytes, Alignment: 64 translationWorker.js:481:21
Aligned memory construction done translationWorker.js:483:21
Aligned memory initialized translationWorker.js:486:21
Aligned vocab memory1 size: 812781 translationWorker.js:343:23
Aligned model memory size: 17140836 translationWorker.js:345:21
Aligned shortlist memory size: 4472528 translationWorker.js:346:21
Translation Model config: 
            beam-size: 1
            normalize: 1.0
            word-penalty: 0
            max-length-break: 128
            mini-batch-words: 1024
            workspace: 128
            max-length-factor: 2.0
[...]

PR #61, opened 2022-01-28
Updated model registry to 0.2.11 version

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/60

It also fixes a part of the problem in https://github.com/mozilla/firefox-translations/issues/58

Test
1. Download Firefox Nightly (German version)
2. Start extension (latest main branch + this PR applied)
3. Navigate to page: https://andrenatal.com/translations-playground/?lang=pt
4. Infobar correctly displays the page language
5. Click on `Übersetzen`
6. The page is translated to German

Replacing `pt` with `es`, `et`, `en`, `it`, `ru` and `cs` in the link https://andrenatal.com/translations-playground/?lang=pt runs successfully.

Also tested on https://www.bbc.com/portuguese and it works there as well.

### DIFF
--- a/extension/model/modelRegistry.js
+++ b/extension/model/modelRegistry.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-unused-vars */
 /* eslint-disable max-lines */
 
-let modelRegistryRootURL = "https://storage.googleapis.com/bergamot-models-sandbox/0.2.10";
+let modelRegistryRootURL = "https://storage.googleapis.com/bergamot-models-sandbox/0.2.11";
 const modelRegistryRootURLTest = "https://example.com/browser/browser/extensions/translations/test/browser";
 
 const modelRegistry = {
@@ -184,9 +184,9 @@ const modelRegistry = {
     model: {
       name: "model.deen.intgemm.alphas.bin",
       size: 17140837,
-      estimatedCompressedSize: 13091214,
+      estimatedCompressedSize: 12995752,
       expectedSha256Hash:
-        "dda44d87ab0d8ad3b3871122fd3ee385f37878183a8b4ec139cd909531ec5009",
+        "1980225d00dc5645491777accff5b3c9d20b92eff67a25135f1cf8fe2ed8fb8f",
     },
   },
   encs: {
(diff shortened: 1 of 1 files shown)


extension/model/modelRegistry.js
```diff
@@ -215,24 +215,24 @@ const modelRegistry = {
   ende: {
     vocab: {
       name: "vocab.deen.spm",
```
jerinphilip, line 217: Can I convince you for duplication here of the same vocab as `vocab.ende.spm` with an argument of consistency for easy automation in https://github.com/browsermt/bergamot-translator/pull/319?

jerinphilip:
```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://storage.googleapis.com/bergamot-models-sandbox/0.2.10/ende/vocab.ende.spm
```

abhi-agg: Creating `vocab.deen.spm` as `vocab.ende.spm` creates a false sense that the two vocab files are different while for most of the language pairs the vocab file for the pair EN <-> Non-EN is just 1 file (e.g. See [this](https://github.com/browsermt/students/blob/master/deen/download-models.sh#L8) and [this](https://github.com/browsermt/students/blob/master/deen/download-models.sh#L22)).

This also keeps the name consistent with the [students](https://github.com/browsermt/students) repository. We took this conscious decision way back when we set up model repository within mozilla org.

jerinphilip: I'm going to go ahead with something like https://gist.github.com/jerinphilip/f9bc31d70e4201771b7734ab1cfd1205 when it gets to working eventually. Ideally, this should come from Mozilla, who maintains a separate model inventory. If you want your models tested (by means of automation provisioned from elsewhere) you are welcome to provide the equivalent and also use the same JSON in the extension than complain when multiple manual copy-pastes break model and configuration. 

I will update the URL to a link provided by you with the equivalent JSON if that ever happens.

kpu:
Please let @abhi-agg update the model and move on with the extension.  I agree model registries are a good idea but don't block this issue on that, open a new one if you want.  Let's check that the files he has are valid and approve if so.

extension/model/modelRegistry.js
jerinphilip: Marking as resolved.

andrenatal (approved):

PR #66, opened 2022-02-01
Fixed issue where non-html text was treated as html input for translation

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/51

cc @jelmervdl @kpu


kpu (approved):
This is the right thing for now.  As discussed, HTML support will flip this back on.

andrenatal (approved):

PR #99, opened 2022-02-11
Cleanup and removal of dead code from translation-notification-fxtranslations.js and fix the option panel in the infobar

### LINKED ISSUE #40: Review infobar UI and make sure all controls are properly working including Firefox panels

### LINKED ISSUE #59: Remove dead code from `translation-notification-fxtranslations.js`

andrenatal:
fixes #59 fixes #40

### DIFF
--- a/extension/view/js/translation-notification-fxtranslations.js
+++ b/extension/view/js/translation-notification-fxtranslations.js
@@ -30,22 +30,6 @@ window.MozTranslationNotification = class extends MozElements.Notification {
               <label anonid="progress-label" value="" style="padding-left:5px"/>
             </hbox>
           </vbox>
-          <hbox class="translated-box" align="center">
-            <label value="&translation.translatedFrom.label;" style="margin-inline-end: 4px;"/>
-            <label anonid="fromLanguage" style="margin-inline:0;font-weight: bold;"/>
-            <label value="&translation.translatedTo.label;" style="margin-inline:3px;"/>
-            <label anonid="toLanguage" style="margin-inline:0;font-weight: bold;"/>
-            <label value="&translation.translatedToSuffix.label;"/>
-            <button anonid="showOriginal" class="notification-button" label="&translation.showOriginal.button;" oncommand="this.closest('notification').showOriginal();"/>
-            <button anonid="showTranslation" class="notification-button" label="&translation.showTranslation.button;" oncommand="this.closest('notification').showTranslation();"/>
-          </hbox>
-          <hbox class="translation-error" align="center">
-            <label value="&translation.errorTranslating.label;"/>
-            <button class="notification-button" label="&translation.tryAgain.button;" anonid="tryAgain" oncommand="this.closest('notification').translate();"/>
-          </hbox>
-          <vbox class="translation-unavailable" pack="center">
-            <label value="&translation.serviceUnavailable.label;"/>
-          </vbox>
         </deck>
         <spacer flex="1"/>
         <button type="menu" class="notification-button" anonid="options" label="&translation.options.menu;">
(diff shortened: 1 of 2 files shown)


extension/view/js/translation-notification-fxtranslations.js
```diff
@@ -169,24 +135,6 @@ window.MozTranslationNotification = class extends MozElements.Notification {
     return this.querySelector(`[anonid=${anonId}]`);
   }
 
-  fromLanguageChanged() {
-    this.translationNotificationManager.reportInfobarEvent("change_lang");
```
eu9ene, line 173: where do we report these infobar events now?

extension/view/js/translation-notification-fxtranslations.js
```diff
@@ -200,57 +148,13 @@ window.MozTranslationNotification = class extends MozElements.Notification {
...
     this.translationNotificationManager.reportInfobarEvent("closed");
-    this._close();
-  }
-
-  /*
-   * to be called when the infobar should be closed per user's wish
-   * by clicking the Not now button
-   */
-  notNow() {
-    this.translationNotificationManager.reportInfobarEvent("not_now");
```
eu9ene, line 229: same question about these two

extension/view/js/translation-notification-fxtranslations.js
andrenatal: Actually if you look carefully, you'll notice that this function was already part of the dead code that I refer to, and it wasn't being called at all. Maybe we should manually triple check in the telemetry backend if all events are actually being sent (from everywhere and not just here). I attached this function to the dropdown event and is not being called.

extension/view/js/translation-notification-fxtranslations.js
andrenatal: Same situation as above: neither `showOriginal()` neither `notNow()` are being called because although this time they are properly attached to the events in the forms, the panel they are part of won't ever be displayed because these functionalities were removed. 

It's not blatantly clear in the code in this file, since the state is controlled by the translationNotificationManager, but by using the extension is easy to notice that these controls are not ever being dsiplayed (we don't need notNow() since there are already the button to dismiss the infobar at the right end", and showOriginal() is not supported by the parser and actually it doesn't make much sense for me and we probably won't implement.

extension/view/js/translation-notification-fxtranslations.js
```diff
@@ -200,57 +151,13 @@ window.MozTranslationNotification = class extends MozElements.Notification {
...
-  }
-
-  _handleButtonHiding() {
-    const originalShown = this.translation.uiState.originalShown;
-    this._getAnonElt("showOriginal").hidden = originalShown;
-    this._getAnonElt("showTranslation").hidden = !originalShown;
-  }
-
-  showOriginal() {
-    this.translationNotificationManager.reportInfobarEvent("accept_outbound");
```
eu9ene, line 240: what about this one?

eu9ene (commented):
We need to make sure that all infobar related telemetry events are reported correctly after refactoring. You can check this by switching telemetry logging parameter in mediator to true and inspecting logs.

extension/view/js/translation-notification-fxtranslations.js
andrenatal: So this event being here was already a mistake. Not only because this function was never called since the panel which contains the element that calls it is never displayed and is disabled, but also its purpose (if you look at the where it is invoked) is to revert from the translated state to the original state of the page, and has nothing to do with what the `accept_outbound` event entails. It seems a lot of these events were wrongly attached to dead and disabled code already that's never being invoked.

andrenatal:
> We need to make sure that all infobar related telemetry events are reported correctly after refactoring. You can check this by switching telemetry logging parameter in mediator to true and inspecting logs.

It's clear that many telemetry events in this file were attached to dead or non-functional code already, so this PR is diverting from its original purpose to something like "Review all Telemetry events", which I think it deserves its own issue so we cold triple check this is not happening in other parts of the code.

eu9ene (approved):

Issue #30, opened 2022-01-13
Improved language detection


andrenatal:
We should have a more expanded language detection heuristic in terms of content extraction and analysis to determine the page's language instead of traversing just through divs.

andrenatal:
Maybe we could use this https://fasttext.cc/docs/en/language-identification.html

kpu:
According to https://github.com/mozilla/firefox-translations/pull/103#discussion_r805665729 the code `no` is coming out of CLD2 instead of `nb` and `nn`.  This is a tell that the small version of the CLD2 classifier is being used.  The full version has more languages and better accuracy (including for languages other than Norwegian).  See https://github.com/bitextor/warc2text/commit/e2543e4d2086a29c914c564aae885e7dcb888dbf for an example of using the full CLD2 classifier.  
Or use fasttext.

PR #103, opened 2022-02-11
Add is, fa, nb, nn models

### LINKED ISSUE #90: Integrate the Norwegian, Icelandic and Farsi models
@abhi-agg and @eu9ene what you think?

eu9ene:
fixes #90

### DIFF
--- a/extension/controller/experiments/LanguageDetector/api.js
+++ b/extension/controller/experiments/LanguageDetector/api.js
@@ -27,7 +27,16 @@ this.experiment_languageDetector = class extends ExtensionAPI {
         languageDetector: {
           detect: async function detect(str) {
             try {
-              return await LanguageDetector.detectLanguage(str);
+              let lang = await LanguageDetector.detectLanguage(str);
+
+              /*
+               * language detector returns "no" for both Norwegian Nynorsk ("nn") and Norwegian Bokmål ("nb")
+               * let's default to "nb", since we have a better model and
+               * Bokmål is more popular in Norway, about 85-90 % of writing is done in Bokmål.
+               */
+              if (lang.language === "no") lang.language = "nb"
+
+              return lang
             } catch (error) {
               // surface otherwise silent or obscurely reported errors
               console.error(error.message, error.stack);
--- a/extension/controller/translation/translationWorker.js
+++ b/extension/controller/translation/translationWorker.js
@@ -200,11 +200,15 @@ class TranslationHelper {
               await this.constructTranslationService();
               await this.constructTranslationModel(sourceLanguage, targetLanguage);
               if (withOutboundTranslation) {
-                await this.constructTranslationModel(targetLanguage, sourceLanguage);
-                postMessage([
-                    "displayOutboundTranslation",
-                    null
-                ]);
+                  try {
+                    await this.constructTranslationModel(targetLanguage, sourceLanguage);
+                    postMessage([
+                        "displayOutboundTranslation",
+                        null
+                    ]);
+                  } catch (ex) {
+                      console.warn("Error while constructing a reversed model for outbound translation. It might be not supported.", ex)
+                  }
               }
               let finish = Date.now();
               console.log(`Model '${sourceLanguage}${targetLanguage}' successfully constructed. Time taken: ${(finish - start) / 1000} secs`);
--- a/extension/controller/translation/translationWorker.js
+++ b/extension/controller/translation/translationWorker.js
@@ -234,7 +234,7 @@ class TranslationHelper {
             this.engineState = this.ENGINE_STATE.LOADED;
             let notificationMessage = "Automatic Translation enabled";
             if (!isReversedModelLoaded) {
-                notificationMessage += "; Translation of forms disabled (reverse translation is not supported for this language)"
+                notificationMessage += ". Translation of forms is not supported for this language."
             }
             postMessage([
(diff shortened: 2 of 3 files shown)


extension/controller/translation/translationWorker.js
```diff
@@ -200,11 +200,15 @@ class TranslationHelper {
...
-                    null
-                ]);
+                  try {
+                    await this.constructTranslationModel(targetLanguage, sourceLanguage);
+                    postMessage([
+                        "displayOutboundTranslation",
+                        null
+                    ]);
+                  } catch (ex) {
+                      console.warn("Error while constructing a reversed model for outbound translation. It might be not supported.", ex)
```
andrenatal, line 210: We need to surface this and warn the user instead of swallowing the error. I've added support to that a couple of commits ago, please refer to that and try to reproduce here. It's just basically sending another message to the mediator, one line.

andrenatal (changes requested):
Didn't test but looks good with the comment below addressed.

extension/controller/experiments/LanguageDetector/api.js
```diff
@@ -27,7 +27,16 @@ this.experiment_languageDetector = class extends ExtensionAPI {
         languageDetector: {
           detect: async function detect(str) {
             try {
-              return await LanguageDetector.detectLanguage(str);
+              let lang = await LanguageDetector.detectLanguage(str);
+
+              /*
+               * language detector returns "no" for both Norwegian Nynorsk ("nn") and Norwegian Bokmål ("nb")
+               * let's default to "nb", since we have a better model and
+               * Bokmål is more popular in Norway, about 85-90 % of writing is done in Bokmål.
```
abhi-agg, line 35: > Bokmål is more popular in Norway, about 85-90 % of writing is done in Bokmål.

@eu9ene Do you have any evidence to back this up?

kpu: CLD2 ships two models, a small one and a full one.  If you're getting `no` it means you're using the small model.  The full model gives `nn` and `nb`.  Related to #30.  

See https://github.com/bitextor/warc2text/commit/e2543e4d2086a29c914c564aae885e7dcb888dbf for an example of changing which model is used.

andrenatal: Maybe @lars-t-hansen can provide us some tips here?

andrenatal: > > > Bokmål is more popular in Norway, about 85-90 % of writing is done in Bokmål.
> > 
> > 
> > @eu9ene Do you have any evidence to back this up?
> > Plus, we should create an issue for ourselves to not forget about updating this code whenever we integrate a better Language detector.
> 
> I don't have "evidence" but some random internet articles say that Bokmal is more popular like this one https://norwegianacademy.com/nynorsk-or-bokmal/.

Maybe @lars-t-hansen can provide us some tips here?

kpu: Bokmål is the dominant idiom.  Norwegians will tell you that including those in the know like the National Library.  

Regarding the web, just look at the file sizes of monolingual dumps here:
https://data.statmt.org/heafield/classified-fasttext/wide00016/nno/
https://data.statmt.org/heafield/classified-fasttext/wide00016/nor/
Much to my chagrin, the fasttext classifier used for that does `nor` when it should emit `nob`.  I have confirmed with the author of the LID used that it `nor` refers to Bokmål.  

That said, you have a classifier that tells you the difference...

extension/controller/translation/translationWorker.js
```diff
@@ -224,10 +232,15 @@ class TranslationHelper {
               return;
             }
             this.engineState = this.ENGINE_STATE.LOADED;
+            let notificationMessage = "Automatic Translation enabled";
+            if (!isReversedModelLoaded) {
+                notificationMessage += "; Translation of forms disabled (reverse translation is not supported for this language)"
```
andrenatal, line 237: Please, replace this by: `notificationMessage += ". Translation of forms is not supported for this language."`

andrenatal (changes requested):

extension/controller/translation/translationWorker.js
eu9ene: done

andrenatal (approved):
Lgtm with the other issues addressed. Tested with Norwegian and it worked, but didn't test the others languages.

extension/controller/experiments/LanguageDetector/api.js
lars-t-hansen: Bokmål is indeed the dominant written idiom (85-90% sounds about right, possibly even low) so for sheer applicability it would be useful to go after that first, but it would be politically hazardous to recognize bokmål and call it simply "Norwegian".

eu9ene: > Bokmål is indeed the dominant written idiom (85-90% sounds about right, possibly even low) so for sheer applicability it would be useful to go after that first, but it would be politically hazardous to recognize bokmål and call it simply "Norwegian".

Thank you for your input! It is called properly in the infobar UI, the only issue is that our language detector cannot distinguish between two idioms, so we default to Bokmål for now. I hope we will fix it in #30.


Screen Shot 2022-02-15 at 10 07 17 AM

Issue #115, opened 2022-02-21
Review and fix localization issues


andrenatal:
As listed here https://github.com/mozilla/firefox-translations/pull/113 an here https://bugzilla.mozilla.org/show_bug.cgi?id=1754942#c2

PR #113, opened 2022-02-19
Extract strings into localizaton files.

andrenatal:
fixes #101

Issue #21, opened 2021-12-28
Implement Localization utilizing pontoon


andrenatal:

andrenatal:
We'll need to develop the system to retrieve the localized strings from pontoon.

PR #116, opened 2022-02-22
Localization fixes

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/115

Addressed comments:
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811175400
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811225648
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811225983
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811226847
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811227548
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811228171
- [x] https://github.com/mozilla/firefox-translations/pull/113#discussion_r811246589
- [x] https://bugzilla.mozilla.org/show_bug.cgi?id=1754942#c5
    - [x] Removed all folders except "en" and renamed "en" to "en-US"
    - [x] Communicate to Pontoon team which locales (with the codes they use internally) need to be enabled in Pontoon


extension/_locales/en-US/messages.json
```diff
@@ -1,14 +1,14 @@
 {
     "extensionName": {
       "message": "Firefox Translations",
-      "description": "Name of the extension"
+      "description": "Name of the extension. This name should stay the same for all locales."
```
flodolo, line 4:
```suggestion
      "description": "Name of the extension. Do NOT translate, keep the name in English."
```

extension/_locales/en-US/messages.json
```diff
@@ -1,14 +1,14 @@
...
     },
     "extensionDescription": {
       "message": "Translate websites in your browser without using the cloud.",
       "description": "Description of the extension."
     },
     "loadingTranslationEngine": {
-      "message": "Loading Translation Engine",
+      "message": "Loading Translation Engine...",
```
flodolo, line 11:
```suggestion
      "message": "Loading Translation Engine…",
```

Always proper ellipsis character.

extension/_locales/en-US/messages.json
```diff
@@ -68,7 +68,7 @@
       }
     },
     "displayStatisticsMessage": {
-      "message": "Display Statistics",
+      "message": "Display Statistics?",
```
flodolo, line 71: Given the screenshot, I'm not completely sure it needs the question mark, and needs a fix for the comment instead.

flodolo: What happens when you click that command? I assume it opens a page for the statistics.

My suggestion was about the comment (i.e. description): `Command used to show statistics (display is intended as a verb)`.

About the message itself: if I my assumption is correct, I don't think it needs a question mark.

extension/_locales/en-US/messages.json
```diff
@@ -68,7 +68,7 @@
       }
     },
     "displayStatisticsMessage": {
-      "message": "Display Statistics",
+      "message": "Command used to show statistics",
```
flodolo, line 71:
```suggestion
      "message": "Display statistics",
```

extension/_locales/en-US/messages.json
```diff
@@ -68,7 +68,7 @@
...
       "description": "Asks the users if they want to display the translation statistics."
```
flodolo, line 72:
```suggestion
      "description": "Command used to show statistics (display is intended as a verb)."
```

flodolo (commented):
Hopefully it's clearer.

flodolo (approved):

extension/_locales/en-US/messages.json
abhi-agg: > What happens when you click that command? I assume it opens a page for the statistics.

It shows the statistics in the info-bar itself. Attaching screenshot for details.

Screenshot 2022-02-22 at 15 03 22

flodolo: Ah, OK. I was expecting a pop-up, or similar. Then maybe update the comment to `Option to show statistics in the infobar (display is intended as a verb).`?

flodolo:
> Now that we removed all the other language folders, is something still left to be done to address this one? I think no. Right?

Yes, you need to tell me which locales need to be enabled in Pontoon (with the codes we use internally).

extension/_locales/en-US/messages.json
abhi-agg: Yeah, sounds good to me.

abhi-agg:
@andrenatal We are left with 2 tasks now (you can see the description for reference).

I created the PR in `mozilla` fork so that you can also collaborate/push changes for the tasks in case I couldn't fix. Please feel free to push commits on top of this branch 👍🏾

flodolo:
Technically, there are a few more tasks, but I don't think they should be blocking this:
- Set up a system to import translations.
- Have the XPI built as an artifact in the GH action, to make it easier to test.

andrenatal:
Thanks @flodolo 

> Technically, there are a few more tasks, but I don't think they should be blocking this:
> 
>     * Set up a system to import translations.
> 

Maybe we should implement that when we start having entries to pull?

>     * Have the XPI built as an artifact in the GH action, to make it easier to test.

We already build xpi's for every push. Please note the `artifacts` section at the bottom of this previously ran action: https://github.com/mozilla/firefox-translations/actions/runs/1883369331

extension/_locales/en_US/messages.json
```diff
@@ -1,14 +1,14 @@
 {
     "extensionName": {
       "message": "Firefox Translations",
-      "description": "Name of the extension"
+      "description": "Name of the extension. Do NOT translate, keep the name in English."
```
andrenatal, line 4: @flodolo @abhi-agg I originally thought the name of the extension should also be translated to the user's language.

flodolo: For reference, all the other add-ons we have remain in English (e.g. "Firefox Multi-account Containers"). 

But, in the end, it's a product call.

flodolo:
> Thanks @flodolo
> 
> > Technically, there are a few more tasks, but I don't think they should be blocking this:
> > ```
> > * Set up a system to import translations.
> > ```
> 
> Maybe we should implement that when we start having entries to pull?

Implementation, yes, that's why I wrote that it shouldn't be blocking. But we need that system sooner than later.

> > ```
> > * Have the XPI built as an artifact in the GH action, to make it easier to test.
> > ```
> 
> We already build xpi's for every push. Please note the `artifacts` section at the bottom of this previously ran action: https://github.com/mozilla/firefox-translations/actions/runs/1883369331

It's a ZIP file though. Can we package the .xpi inside that ZIP file instead?

extension/_locales/en_US/messages.json
andrenatal: We don't have product support at the moment, so let's follow the other add-ons policy then.

andrenatal:
> > Thanks @flodolo
> > > Technically, there are a few more tasks, but I don't think they should be blocking this:
> > > ```
> > > * Set up a system to import translations.
> > > ```
> > 
> > 
> > Maybe we should implement that when we start having entries to pull?
> 
> Implementation, yes, that's why I wrote that it shouldn't be blocking. But we need that system sooner than later.
> 
> > > ```
> > > * Have the XPI built as an artifact in the GH action, to make it easier to test.
> > > ```
> > 
> > 
> > We already build xpi's for every push. Please note the `artifacts` section at the bottom of this previously ran action: https://github.com/mozilla/firefox-translations/actions/runs/1883369331
> 
> It's a ZIP file though. Can we package the .xpi inside that ZIP file instead?

We can't since that's a limitation of the upload api, but I don't see how that's a problem, since the zip file can be used to install the addon via `about:debugging` like the xpi, which, in the end, is a zip file too.
```

MacBookPro:web-ext-artifacts anatal$ file -I firefox_translations.xpi
firefox_translations.xpi: application/zip; charset=binary
```

flodolo:
> We can't since that's a limitation of the upload api, but I don't see how that's a problem, since the zip file can be used to install the addon via `about:debugging` like the xpi, which, in the end, is a zip file too.

Because most localizers are not technical. The simpler we can make their life, the better.

To be clear: I'm not asking to upload the .xpi directly as an artifact, but a .zip of the .xpi. MAC is [currently doing](https://bugzilla.mozilla.org/show_bug.cgi?id=1734220) that (don't know the details of how they do it).

flodolo:
> Because most localizers are not technical. The simpler we can make their life, the better.

For reference, these are [the instructions](https://docs.google.com/document/u/1/d/e/2PACX-1vQi-OxGobQbb9VLvI_9HTBN0Ed_0FOJleBHKGyBCeDJICx_58iaxcdPuqVj0I9rR7SkuwDZYnpgp85Z/pub) I wrote for MAC. Going through about:debugging is probably a bit too much, so it would be great if we could skip that.

andrenatal:
Thanks @flodolo . We'll try to change that here: https://github.com/mozilla/firefox-translations/pull/117#pullrequestreview-890331885, and then we modify the documentation in the README to simplify it like you suggested. 

Re: system to pull the strings: That's planned to happen on next week (https://github.com/mozilla/firefox-translations/issues/21), but if you say we'll have strings coming before that, I can prioritize that and do it as soon we have them. As I mentioned, we just need to know from where to to pull.

Re:
> Yes, you need to tell me which locales need to be enabled in Pontoon (with the codes we use internally).

I couldn't find where all these codes are located internally in the repo neither in the doc but I might have missed it, could you please point to where is the reference?

andrenatal:
I'll land this here and followup about the locales in the bmo bug since is more appropriate than here.

andrenatal (approved):

extension/_locales/en_US/messages.json
abhi-agg: Resolving this one.

flodolo:
> @flodolo I couldn't find any step [here](https://github.com/mozilla/firefox-translations/#testing) that involves `about:debugging`. Did you mean `about:addons` step can be avoided?

That's my fault. I thought you could only install from an `.xpi`, not from `.zip` in `about:addons`. Turns out I was wrong, and it even works if you drag and drop the `.zip` file on `about:addons` (but not on other windows).

Given this, I think we're good without doing extra work to change it to `.xpi`

> Why should the zipped .xpi be created in the artifacts instead of just .xpi when we know that the user has to finally install the .xpi?

The suggestion was only to avoid Firefox trying to install directly from GitHub when clicking the link.

abhi-agg:
Cool. Then we don't need to do anything in https://github.com/mozilla/firefox-translations/pull/117. Thanks a lot for clarification 👍🏾

PR #121, opened 2022-02-23
Adding _locales submodule pointing to https://github.com/mozilla-l10n/firefox-translations-l10n

### LINKED ISSUE #21: Implement Localization utilizing pontoon

andrenatal:
Fixes #21 

This is inspired on the same approach that https://github.com/mozilla-l10n/multi-account-containers-l10n/ uses. 

I like that because then we won't need actions or bots to push PRs to keep the `_locales` folder synced, and we can just utilize npm runners when we need it updated to both develop as to create builds both manually as on CI.


package.json
```diff
@@ -3,12 +3,13 @@
...
-    "format": "prettier 'extension/*.{js,css}' --tab-width=2 --arrow-parens=always --trailing-comma=es5 --no-bracket-spacing --write",
+    "build": "npm run update-localization; web-ext build -s extension --overwrite-dest  --no-config-discovery -n firefox_translations.xpi",
     "lint:extension": "web-ext lint -s extension",
-    "once": "web-ext run -s extension --verbose --firefox nightly  --no-config-discovery --pref=extensions.experiments.enabled=true --pref=extensions.translations.disabled=false --pref=javascript.options.wasm_simd_wormhole=true --pref=datareporting.healthreport.uploadEnabled=false",
+    "once": "npm run update-localization; web-ext run -s extension --verbose --firefox nightly  --no-config-discovery --pref=extensions.experiments.enabled=true --pref=extensions.translations.disabled=false --pref=javascript.options.wasm_simd_wormhole=true --pref=datareporting.healthreport.uploadEnabled=false",
     "package": "npm run build && mv web-ext-artifacts/*.zip addon.xpi",
-    "setup-webext": "npm install -g web-ext"
+    "setup-webext": "npm install -g web-ext",
+    "clean-localization-folder": "rm -rf extension/_locales/.github",
```
andrenatal, line 11: we need this because `web-ext` fails to load the extension when there's an extraneous folder inside `_locales`

flodolo (approved):
If this is the approach, we can remove the CODEOWNERS file.

The downside is developing new features or changes that require changes to the en-US file (you'll need a PR to land them in the l10n repository before merging them).

andrenatal:
> If this is the approach, we can remove the CODEOWNERS file.
> 

cool, just removed it 

> The downside is developing new features or changes that require changes to the en-US file (you'll need a PR to land them in the l10n repository before merging them).

we can live with that, don't think there will be too many changes on that

PR #123, opened 2022-02-23
Fixed README to avoid failures while installing the Extension

abhi-agg:
~~- Corrected the value for the pref `extensions.translations.disabled`~~
~~- Should be set to `false` for using the extension~~

EDIT: Updating the description of the PR

Fixes a part of https://github.com/mozilla/firefox-translations/issues/125 (as per https://github.com/mozilla/firefox-translations/issues/125#issuecomment-1049889063)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ One can test preliminary versions of the extension in Firefox Nightly since an a
 - Download [Firefox Nightly](https://www.mozilla.org/en-US/firefox/all/#product-desktop-nightly)
 - Type `about:addons` in the navigation bar, click on the gear to he right, then on `install from a file` and load the saved zip file.
 - Type `about:config` in the navigation bar again, and switch the following preferences:
-  * `extensions.translations.disabled` to `true`
+  * `extensions.translations.disabled` to `false`
   * `extensions.experiments.enabled` to `true` 
   * `javascript.options.wasm_simd_wormhole` to `true`
   * `xpinstall.signatures.required` to `false`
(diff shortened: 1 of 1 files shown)


andrenatal:
@abhi-agg did you have problems following this step by step? The problem of having this set to true is that the legacy integrated version would be loaded. When we install the addon manually in Nightly following those steps, both will be installed at same time, and it you have the pref to false, both will be enabled at same time too.

abhi-agg:
@andrenatal Yes. I followed the steps and I see the same behavior as @flodolo observed.

Setup was latest FFx Nightly (en-US). I installed the [latest extension from main branch](https://github.com/mozilla/firefox-translations/actions/runs/1886435988) and opened a spanish website `https://www.bbc.com/mundo`. The infobar never appeared with [these](https://github.com/mozilla/firefox-translations#testing) instructions.

However, after setting `extensions.translations.disabled` to `false`, the infobar started to appear. I am assuming that the extension that is running now is the one that I loaded from file because I see Outbound and QE specific check-boxes in the infobar which weren't a part of the in-tree extension. Please correct me here 👍🏾

Attaching screenshot for details.

Screenshot 2022-02-24 at 12 10 52

andrenatal (approved):

flodolo:
Once piece of my comments that hasn't been captured, and still happens: following the instructions from the README works, but every time I restart Firefox the add-on gets disabled (I assume because of the `extensions.translations.disabled` key).

Issue #52, opened 2022-01-26
Use HTML translation feature including inline tag movement


kpu:
Currently the code sends text snippets for translation:

https://github.com/mozilla/firefox-translations/blob/c21b61fdd1dc3186bb478c7702d601dc20b7b336/extension/view/js/InPageTranslation.js#L135-L158

This leads to very poor translation quality because the system does not have sentence context.  Even if it were to have sentence context, keeping text spans as is prevents reordering and is an impossible translation problem.  For example, `chien` translates to `dog`.  In this HTML, what is the translation of `h`?  
`<span id="0">c</span><span id="1">h</span><span id="2">i</span><span id="3">e</span><span id="4">n</span>` 

Since block elements are sentence-breaking, individual block elements can be sent for translation using their `innerHTML`.  The HTML parser also knows to break sentences at block boundaries so larger elements can also be sent in.  It does assume well-formed HTML though; Firefox is better at fixing HTML and this ensures consistency between rendering and how the engine perceives tags.  Well-formed implies tags that open also close inside the same block of text; #23 is a blocker.  

#51 is a partial blocker.  Specifically this part needs to be fixed first:

> Even if HTML was being submitted, it would not be properly used (and cause an `abort()`) because the model doesn't produce alignment information. In [the model configuration yaml](https://github.com/mozilla/firefox-translations/blob/c21b61fdd1dc3186bb478c7702d601dc20b7b336/extension/controller/translation/translationWorker.js#L262-L275), the line `alignment: soft` is missing.

Once that is fixed, HTML processing coming out of the engine should be consistent with https://translate.ikhoefgeen.nl/ .  

Quality issues with HTML processing should be raised on https://github.com/browsermt/bergamot-translator

andrenatal:
The team has very little confidence in the current bergamot-translator's embedded HTML translation capabilities due the very well documented issues and long time it took to have it implemented along the constant belittling displayed from your team in regards to the way we were trying to solve this problem while your team was still unable to provide the right tools (again also well documented).

But we decided to give it another try in a couple weeks and submit it to QA, but if we'll still have issues like page defacing, stripping of htmls tags and etc, we will abandon this altogether and remain with textNodes which will be the approach utilized in the user test we are mandated to run internally.

abhi-agg:
https://github.com/mozilla/firefox-translations/pull/111 imported the latest API changes in bergamot-translator which will enable the extension to use HTML translation feature.

[...]

Issue #23, opened 2021-12-28
Properly split the content when parsing html to respect  opening and closing tags


andrenatal:

andrenatal:
Fixed by #127

PR #127, opened 2022-02-25
Adding HTML Translations

andrenatal:
@abhi-agg @eu9ene I was very happy to find that Jelmer fork https://github.com/jelmervdl/firefox-translations/blob/v0.2.6/extension/view/js/InPageTranslation.js utilized and developed further the same approach we discussed and had sitting in the `html` branch in this repo. 

I reviewed and tested the code and it's looking great! There was some issues but fixable. Let's continue from there. 

Thanks @jelmervdl!

### DIFF
--- a/extension/controller/translation/Translation.js
+++ b/extension/controller/translation/Translation.js
@@ -14,6 +14,7 @@ class Translation {
         this.translateSchedule = null; // holds a reference to the translation setTimeout
         this.translationMessageBuffer = new Queue();
         this.mediator = mediator;
+        this.htmlRegex = new RegExp("<(.*)>.*?|<(.*) />", "gi");
         const engineLocalPath = browser.runtime.getURL("controller/translation/bergamot-translator-worker.js");
         const engineRemoteRegistry = browser.runtime.getURL("model/engineRegistry.js");
         const modelRegistry = browser.runtime.getURL("model/modelRegistry.js");
@@ -149,6 +150,8 @@ class Translation {
         this.translationsMessagesCounter += 1;
         translationMessage.messageID = this.translationsMessagesCounter;
         translationMessage.sourceParagraph = sourceParagraph;
+        // let's revisit this later, since passing false here when there's plain text is breaking
+        translationMessage.isHTML = true; // this.htmlRegex.test(sourceParagraph);
         switch (type) {
             case "outbound":
                 translationMessage.sourceLanguage = navigatorLanguage;
--- a/extension/controller/translation/translationWorker.js
+++ b/extension/controller/translation/translationWorker.js
@@ -573,7 +573,7 @@ class TranslationHelper {
                 vectorResponseOptions.push_back({
                     qualityScores: false,
                     alignment: true,
-                    html: false,
+                    html: true,
                 });
             });
             if (vectorResponseOptions.size() === 0) {
--- a/extension/controller/translation/translationWorker.js
+++ b/extension/controller/translation/translationWorker.js
@@ -573,7 +573,7 @@ class TranslationHelper {
                 vectorResponseOptions.push_back({
                     qualityScores: false,
                     alignment: true,
-                    html: true,
+                    html: message.isHTML,
                 });
             });
             if (vectorResponseOptions.size() === 0) {
--- a/extension/view/js/InPageTranslation.js
+++ b/extension/view/js/InPageTranslation.js
@@ -1,22 +1,15 @@
 /*
- * extracted from the fork
- * https://github.com/jelmervdl/firefox-translations/blob/v0.2.6/extension/view/js/InPageTranslation.js
+ * this is the DOM Parser responsible for both walk as observe for mutations
+ * in the tree and submit translations and also render them when returned
+ * authors: @andrenatal, @jelmervdl
  */
 
-"use strict";
-
-function computePath(node, root) {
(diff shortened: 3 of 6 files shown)


andrenatal:
Fixes #52 and #23

andrenatal:
@abhi-agg @eu9ene lmk what you think. If this sounds good to you I'll make the texts and linter pass.  @jelmervdl don't refrain of contributing back your changes here!

extension/controller/translation/translationWorker.js
```diff
@@ -573,7 +573,7 @@ class TranslationHelper {
                 vectorResponseOptions.push_back({
                     qualityScores: false,
                     alignment: true,
-                    html: false,
+                    html: true,
```
abhi-agg, line 576: @andrenatal We should set this flag programmatically. It should be passed from the files in `view` (based on whether the text to be translated is html or plain) to the mediator and then should reach here so that the responseOptions can be constructed per item of the batch.

extension/view/js/InPageTranslation.js
```diff
@@ -112,21 +311,24 @@ class InPageTranslation {
          */
         this.translationsCounter += 1;
 
+        // Debugging: mark the node so we can add CSS to see them
+        node.setAttribute("x-bergamot-translated", this.translationsCounter);
```
abhi-agg, line 315: Do we need to add CSS for the case when QE is not selected by user? Based on my understanding, this code is written assuming QE is always on. We need to refactor the whole file to mark the nodes to add CSS only when QE is on.

extension/view/js/InPageTranslation.js
```diff
@@ -55,22 +150,31 @@ class InPageTranslation {
...
+    addDebugStylesheet() {
+        const element = document.createElement("style");
+        document.head.appendChild(element);
+
+        const sheet = element.sheet;
+        sheet.insertRule("html[x-bergamot-debug] [x-bergamot-translated] { border: 2px solid red; }", 0);
+        sheet.insertRule("html[x-bergamot-debug] [x-bergamot-translated~=\"skipped\"] { border: 2px solid purple; }", 1);
+        sheet.insertRule("html[x-bergamot-debug] [x-bergamot-translated~=\"rejected\"] { border: 2px solid yellow; }", 2);
+        sheet.insertRule("html[x-bergamot-debug] [x-bergamot-translated=\"\"] { border: 2px solid blue; }", 3);
+        sheet.insertRule("html[x-bergamot-debug] [x-bergamot-translated=\"\"] [x-bergamot-translated~=\"is-excluded-node\"] { border: 4px dashed red; }", 4);
```
abhi-agg, line 162: What does `x-bergamot-debug` and `x-bergamot-translated` mean here? What is the purpose of this whole function? Is it usable in case of QE only?

extension/view/js/InPageTranslation.js
jerinphilip: This is debugging CSS. Please use the extension.

image

abhi-agg: Which extension are you talking about?

extension/view/js/InPageTranslation.js
jelmervdl: You can safely remove all getAttribute and setAttribute code. Its only there to help with identifying what blocks get send for translation, or why they get skipped. Nothing of this is related to QE.

The only functional usage of attributes is the `node.dataset.xBergamotId` and `data-x-bergamot-id` bits that are used to merge tags.

abhi-agg: Thanks for clarification @jelmervdl 👍🏾

extension/view/js/InPageTranslation.js
jelmervdl: https://github.com/jelmervdl/firefox-translations/releases/tag/v0.2.6 ([source code](https://github.com/jelmervdl/firefox-translations))

extension/view/js/InPageTranslation.js
abhi-agg: Btw, I couldn't find any reference to `data-x-bergamot-id` in the code.

jelmervdl: That's correct. `node.dataset.xBergamotId` [sets that attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) on the nodes when they are submitted to the wasm worker. Only during that time do the attributes exist. You should be able to see them at that moment in the DOM tree of the document as well as in the HTML string that's submitted to the WASM worker. I think they currently linger around after that, but in theory they can be removed again at the end of `merge()`.

andrenatal:
I'll spend tomorrow morning working to fix this and make it pass.

extension/controller/translation/Translation.js
```diff
@@ -149,6 +150,8 @@ class Translation {
         this.translationsMessagesCounter += 1;
         translationMessage.messageID = this.translationsMessagesCounter;
         translationMessage.sourceParagraph = sourceParagraph;
+        // let's revisit this later, since passing false here when there's plain text is breaking
+        translationMessage.isHTML = true; // this.htmlRegex.test(sourceParagraph);
```
abhi-agg, line 154: As proposed by @andrenatal, we will figure out this part together and in a separate issue to unblock QA.

abhi-agg (approved):

Issue #133, opened 2022-03-02
Switch Quality Estimation specific model config off when not in use in order to have better performance


abhi-agg:
Whenever QE is off, construct the models with `skip-cost` set to `true`.

andrenatal:
Closed by #144

Issue #132, opened 2022-03-02
Integrate Basic Quality Estimation


abhi-agg:
Basic QE means: Using the free quality scores coming from the translation models

- [ ] Update bergamot translator engine (if required)
- [ ] Update the translation worker with the corresponding API changes

andrenatal:
Closed by #144

Issue #96, opened 2022-02-10
Use translation cache


jerinphilip:
The translation engine provides a cache (https://github.com/browsermt/bergamot-translator/issues/201) at the sentence level. In addition, there is some experimentation and study from https://github.com/XapaJIaMnu/translateLocally/pull/75. 

For the extension, we expect this to be useful at least in the following scenarios:

1. **Recurring page elements over multiple pages** Pages like Wikipedia or news websites which users browse continuously and have some recurring across pages text (e.g. sidebars and such UI elements) that will be repeatedly translated each time. Translation (and the underlying matrix multiplications) are time-consuming. The cache would avoid some heavy computations that would occur otherwise. 
2. **Outbound translation** If the user is entering multiple sentences, then repeatedly translating the whole sentences over and over will lead to sluggishness. I experience this when I type a large text and continue to edit it at https://mozilla.github.io/translate/. The updates on the right happen for about 20 seconds after I type in, even at say 500 characters.

The cache is easy to enable - implement the bindings for cache config (two values - how many sentences to store, whether to enable or not) at WebAssembly and start configuring at JS. Please consider using the cache.

andrenatal:
The extension creates one engine per page load in its own translationWorker, which is not shared across other tabs, so the state is not shared across multiple pages/reloads. How it would work in this case?

andrenatal:
Or better: if there's a way to enable is only when OT is in use, then it could be fine.

andrenatal:
Reminder to self: revisit with the secops team if it's ok to host the nmt wasm module in the background script, which would make translation happen faster. Also assess the memory usage implications of this approach.

andrenatal:
Closed by #144

Issue #26, opened 2022-01-11
[meta] Implement Basic Quality Estimation


andrenatal:
We need to be able to enable the quality estimation functionality by switching a pref. This same pref should also disable the functionality that causes performance penalty in the engine, and not only the UI feature.

List of tasks:
- [x] #132
- [x] Modify extension to encode plain text and send it as HTML to the engine whenever QE feature is on
    - This is required as engine always expects html text to return QE scores (as per [this](https://github.com/browsermt/bergamot-translator/pull/358#issuecomment-1050692865))
    - The translation doesn't need to be decoded back as the engine always returns translation as html whenever QE is on and this output can be used directly to show colors
- [x] #133
- [x] #179

andrenatal:
Closing this since all references items were closed.

PR #144, opened 2022-03-07
Integrate Basic QE feature

abhi-agg:
Fixes:
https://github.com/mozilla/firefox-translations/issues/133
https://github.com/mozilla/firefox-translations/issues/132
https://github.com/mozilla/firefox-translations/issues/96


abhi-agg:
> - [x] #132
> - [x] Modify extension to encode plain text and send it as HTML to the engine whenever QE feature is on
>     - This is required as engine always expects html text to return QE scores (as per [this](https://github.com/browsermt/bergamot-translator/pull/358#issuecomment-1050692865))
>     - The translation doesn't need to be decoded back as the engine always returns translation as html whenever QE is on and this output can be used directly to show colors
> - [x] #133
> - [ ] Implement a mechanism to show colors for in-page translation from the quality scores returned by the engine
>     - [Quality scores are embedded in translation result in form of font attributes](https://github.com/browsermt/bergamot-translator/pull/358#issue-1142083849)

Fixes first 3 parts of https://github.com/mozilla/firefox-translations/issues/26. The last part is left which will be taken up the next week.

andrenatal (approved):

extension/controller/translation/translationWorker.js
```diff
@@ -246,7 +288,7 @@ class TranslationHelper {
         // instantiate the Translation Service
         constructTranslationService() {
             if (!this.translationService) {
-                let translationServiceConfig = {};
+                let translationServiceConfig = { cacheSize: 10 };
```
jelmervdl, line 291: Just a note: a cache size of 10 won't do you much good. The cache is non-probing, [it is basically just `cache[hash(sentence) % cacheSize]`](https://github.com/browsermt/bergamot-translator/blob/22d6bc07e7a6e83bc22676a106fe0ae8fc691c15/src/translator/cache.h#L65-L70). The [description of the parameter](https://github.com/browsermt/bergamot-translator/blob/22d6bc07e7a6e83bc22676a106fe0ae8fc691c15/src/translator/service.h#L33-L37) is a bit indirect about that. If you set it too low, you'll end up having too many different sentences hitting the same cache entry, constantly overwriting each other, and no cache benefit at all.

In my experience you'd get about 20% occupancy, so if you set it to 50 you'd be caching about 10 sentences. But from testing in TranslateLocally and my extension fork, I'd suggest starting with something around 1000 or higher, and see whether you can notice it in the memory usage.

andrenatal: Thanks for the tip Jelmer, please keep them coming. I think caching will be particularly helpful after I move the engine to the background script last week. I heard from the security folks that's fine, so I think that will be helpful.

Issue #135, opened 2022-03-02
Load the wasm module in the background script as a singleton


andrenatal:

andrenatal:
I experimented with this and it seems *slightly* faster since we are cutting off the model loading time (the engine loading is irrelevant, it takes 0.07 seconds for me). I could not get a hold of anyone from security to review this since it seems everyone is, understandably, busy with more important matters, so let's move this to next milestone when, hopefully, the official addon review begins.

andrenatal:
I explored this further and reached a working point, but after having ~6 models loaded, the engine started to run into memory allocation issues which is probably a restriction in the Web Extensions API. Although there's some speedup to start the translation when the model is already loaded, the costs and risks outweigh the gains, specially when one needs to have outbound translation and quality estimation models loaded at the same time for many languages. So let's abandon this and keep one engine being lazy loaded per tab and find a way to route the translations requests from iframes to the worker via the bgscript and the mediator to solve #87 

```
[2022-03-16 14:16:29] Error: Failed to allocate memory on CPU bergamot-translator-worker.js:1217:12
[2022-03-16 14:16:29] Error: Aborted from void *marian::cpu::(anonymous namespace)::genericMalloc(size_t, size_t) in /root/checkout/3rd_party/marian-dev/src/tensors/cpu/aligned.h:33 bergamot-translator-worker.js:1217:12
Callstacks not supported in WASM builds currently
```

andrenatal:
After more discussions with the webextensions team about this, we found there's not any type of memory limitations for the extension process, but they advised against keeping long lived objects with large amounts of memory allocated in the background script for a number of reasons that could be detrimental to performance.

PR #148, opened 2022-03-10
Iframe support

### LINKED ISSUE #87: Yandex feedback form is not translated
1. Go to https://yandex.ru/support/connect-forms/common/feedback.html
2. Press translate with "Enable translation of forms" checked.
3. Click the form text field

Expected: 
1. The page is completely translated from Russian to English
2. Outbound translations dialogue appears

Actual:
1. The page is translated only partially and the form is not translated at all
2. Outbound translations dialogue doesn't appear

Screen Shot 2022-02-09 at 11 50 45 AM

eu9ene:
- fixes #87 
- enable loading of content scripts in all frames
- add conditions do some actions only in the top frame
- move telemetry to background script

Side effects: we still have the translation worker in the content script, which means the wasm module and the models will be loaded for every frame.

AFAIK @andrenatal is working on moving some of them to bgScript


andrenatal:
Just reviewed this, and have few concerns:
- We can't load one module per frame, even temporarily. If a page has 10 frames, or constructs them dynamically, that will likely crash the tab if not the browser. Let's wait for #135 to be landed next week then and refactor this to adapt to the single engine approach.
- There's a ton of different changes and this touches a lot of code that not always seemed related to the PR, e.g. changes to package.json, telemetry, code styling and etc that at some point I didn't even know what I was reviewing anymore. This will need to be broken down in smaller pieces 
- I didn't understand the reasoning behind the change to the name of the messages 
- There is a lot of changes to spaces, either adding or removing. 
- We'll need to have test coverage for frames too.

andrenatal (changes requested):
Let's wait until we have the singleton engine available. Multiple modules per page isn't acceptable.

eu9ene:
>We can't load one module per frame, even temporarily. If a page has 10 frames, or constructs them dynamically, that will likely crash the tab if not the browser. Let's wait for https://github.com/mozilla/firefox-translations/issues/135 to be landed next week then and refactor this to adapt to the single engine approach.

Sure, let's wait

> There's a ton of different changes and this touches a lot of code that not always seemed related to the PR, e.g. changes to package.json, telemetry, code styling and etc that at some point I didn't even know what I was reviewing anymore. This will need to be broken down in smaller pieces


This can't be merged without fixing telemetry because it will mess up a lot of metrics. 

> I didn't understand the reasoning behind the change to the name of the messages

This all has become more complex, so I tried to unify names of messages that are involved in reporting of telemetry to simplify reading of code 

>There is a lot of changes to spaces, either adding or removing.

Sorry about styling, this is IntelliJ autoformatting, I'll see what I can do about it

> We'll need to have test coverage for frames too

Good point, I haven't added any test cases so far. It might take some time though.

eu9ene:
@andrenatal the following improvements were implemented:
1. Iframes don't load translation workers anymore, only the top frame performs translation, other frames communicate with the top frame through background script. InPageTranslation and OutboundTranslation classes still work inside every frame because they need direct access to DOM. The only downside of this is that form translation UI is shown inside iframe. It looks fine to me and I think this use case is not very popular.
2. added a test for iframe
3. fixed some whitespace issues

andrenatal:
@eu9ene Did you ever tested it outside the yandex page in question? I did and although it's working, it introduces something that makes the browser request a confirmation due changes to the page to every navigation. You don't even need to push the translation button: just having the the infobar displayed and navigating to any page causes the error below: 

Screen Shot 2022-03-21 at 8 13 08 PM

andrenatal (changes requested):
the issue with the confirmation due changes to the page being displayed to every navigation should be fixed

eu9ene:
Yes, I tested on some other websites and I see that this doesn't happen on Wikipedia for example. I saw this dialogue on a page with a form and thought it was a consequence of the form not being filled. Now I see that this happens on habr.ru too without forms. Let me investigate this. And thanks for testing!

andrenatal (approved):

Issue #107, opened 2022-02-16
Add telemetry for Quality Estimation.


andrenatal:
@abhi-agg @eu9ene Do we need any for that? At least report when the user selects the checkbox in the infobar would make sense.

eu9ene:
We need something to filter on (activation event or boolean for the checkbox) and there is a string field `performance.translation_quality` to save some values. We can add another field if we want to save numbers.

eu9ene:
I see that the quality estimator fills score attributes for every sentence and every word. Obviously, we can't report raw data, so I propose reporting **average**, **median** and **90th percentile** for word and sentence scores per translated page (total 6 metrics). Recording distributions with more buckets would be nice, but it's more work especially on the analytics side because telemetry doesn't support custom distributions. 

Also, quality estimation modifies HTML significantly which might lead to a slow-down of user experience and there is no benefit for a user to enable it, so I expect very few data points.

PR #161, opened 2022-03-15
Add metrics for quality estimation

eu9ene:
Metrics are added according to the proposal in #107 

The next step is to implement collection of metrics that includes:
- extracting scores from html attributes somewhere
- pass word and sentence scores to mediator using messaging
- call telemetry.addQualityEstimation ( it can be called many times, for every translation batch)

### DIFF
--- a/extension/model/telemetry/Telemetry.js
+++ b/extension/model/telemetry/Telemetry.js
@@ -61,7 +61,9 @@ class Telemetry {
         this._updateUsageTime();
     }
 
-    addQualityEstimation(wordScores, sentScores) {
+    addQualityEstimation(wordScores, sentScores, isSupervised) {
+        this._client.boolean("quality", "is_supervised", isSupervised);
+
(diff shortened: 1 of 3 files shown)


jelmervdl:
This might be relevant: In [bergamot-translator#370](https://github.com/browsermt/bergamot-translator/pull/370) it appears  that the quality scores are not linear, and the scores from the unsupervised models (I don't know which ones those are) may not be comparable to each other.

eu9ene:
> quality scores are not linear

median and 90th percentile will still be useful to approximate distribution and compare models/pages between each other.

> the scores from the unsupervised models (I don't know which ones those are) may not be comparable to each other

if this is the case we can add a boolean metric like  "quality_supervised" to be able to filter later and analyze those types of models separately

andrenatal:
I'll defer this to @abhi-agg who understands this stuff better

andrenatal (approved):

PR #186, opened 2022-03-26
Resubmit textNode parent when one is found in the walker

### LINKED ISSUE #183: Astray textNodes on mutations are breaking in-page translation
@jcchuang reported that on https://www.onlyu.cz/, after the translation has started, any resize of the window will re-render the cookie consent modal, and, consequently replace the elements with its original version. So the problem is that in that mutation, there's a textNode that's finding its way through the treewalker, consequently breaking the rest of the flow, which actually expects elements. I have a patch for that.

andrenatal:
fixes #183 
this did the trick. @jelmervdl what you think?

### DIFF
--- a/extension/view/js/InPageTranslation.js
+++ b/extension/view/js/InPageTranslation.js
@@ -176,6 +176,12 @@ class InPageTranslation {
 
     startTreeWalker(root) {
 
+        // if we have a textNode we walk through its parent and reject it
+        if (root.nodeType === 3){
+            this.startTreeWalker(root.parentNode);
+            return;
+        }
+
         /*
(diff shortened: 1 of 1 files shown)


extension/view/js/InPageTranslation.js
```diff
@@ -176,6 +176,12 @@ class InPageTranslation {
 
     startTreeWalker(root) {
 
+        // if we have a textNode we walk through its parent and reject it
+        if (root.nodeType === 3){
```
eu9ene, line 180: it would be good to introduce named constants for nodeType 1 and 3, it's not clear right away what they mean

eu9ene (approved):

PR #187, opened 2022-03-26
Improve the displaying of the translaton infobar

andrenatal:
fixes #178

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 /* global LanguageDetection, browser, PingSender, BERGAMOT_VERSION_FULL, Telemetry, loadFastText, FastText */
 
 /*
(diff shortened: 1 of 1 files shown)


andrenatal:
Continuation of https://github.com/mozilla/firefox-translations/pull/178

extension/controller/backgroundScript.js
```diff
@@ -80,8 +81,20 @@ const messageListener = async function(message, sender) {
...
+                            tab.id,
+                            { command: "responseMonitorTabLoad", tabId: tab.id }
+                        );
+                    } ,250);
+                } else if (tab.id !== sender.tab.id) {
+                    browser.tabs.onUpdated.removeListener(listeneronUpdatedLoad);
+                    browser.webNavigation.onCompleted.removeListener(webNavigationCompletedLoad);
+                    console.log("browser.tabs.onUpdated.addListener => notifying browser to display the infobar:  tab.id !== sender.tab.id", changeInfo.status, tab.id, sender.tab.id, tab.url)
+
+                    setTimeout(() => {
```
Rob--W, line 93: This `setTimeout` logic is very unreliable. There are several ways to fail with this logic, e.g. page navigated (or did not finish navigation), listener not registered yet, tab closed, etc.

It looks like you're trying to detect the navigation to the next tab and then sending the tabId to the next page, hoping to reach the content script. If all that you're interested in is the tabId, then you can call `browser.runtime.sendMessage` from the content script and return the value from the `browser.runtime.onMessage` listener.

If you're not just interested in the tabId, but also some state from the previous page, then you could maintain a global `Map` in your background script, where you associate relevant information with the tab. If you wish, you could lazily register and unregister tab event listeners to update or expire entries in that `Map` (i.e. if the map is empty, unregister these events, but if the map was empty but not any more, register these tabs listeners).

Your "wait until the page within the tab is loaded" logic assumes that the event happens. This assumption is invalid: the navigation could fail, or the tab could be closed. To detect closed tabs, use the `browser.tabs.onRemoved` event.

extension/controller/backgroundScript.js
```diff
@@ -80,8 +81,20 @@ const messageListener = async function(message, sender) {
                      */
                     setTimeout(() => {
                         browser.tabs.sendMessage(
-                            tabId,
-                            { command: "responseMonitorTabLoad", tabId }
+                            tab.id,
+                            { command: "responseMonitorTabLoad", tabId: tab.id }
+                        );
+                    } ,250);
+                } else if (tab.id !== sender.tab.id) {
```
Rob--W, line 88: This logic would activate the logic for an arbitrary tab (that happens to trigger the `tabs.onUpdated` event). Is that really what you intended?

Issue #134, opened 2022-03-02
Integrate Advanced Quality Estimation


abhi-agg:
Use supervised QE models to get Quality Scores

- [x] Integrate [Supervised QE binary models](https://github.com/browsermt/quality-estimation) in bergamot-translator
    - [x] https://github.com/browsermt/bergamot-translator/issues/318
    - [x] Update test page with supervised QE models specific JS bindings
- [ ] Update bergamot translator engine (if required) in the extension
- [ ] Update the translation worker with the corresponding API changes
- [x] Add Supervised QE binary models in firefox-translations-models repo (https://github.com/mozilla/firefox-translations-models/issues/37)
- [x] Update the `firefox-translations-models` version to use the Supervised QE models in the extension

abhi-agg:
> - [x] Add Supervised QE binary models in firefox-translations-models repo (https://github.com/mozilla/firefox-translations-models/issues/37)

Done by https://github.com/mozilla/firefox-translations-models/pull/38

abhi-agg:
> - [x] Integrate [Supervised QE binary models](https://github.com/browsermt/quality-estimation) in bergamot-translator
>     - [x] https://github.com/browsermt/bergamot-translator/issues/318
>     - [x] Update test page with supervised QE models specific JS bindings

Done by https://github.com/browsermt/bergamot-translator/pull/378

abhi-agg:
> - [x]  Update the `firefox-translations-models` version to use the Supervised QE models in the extension

Done by https://github.com/mozilla/firefox-translations/pull/189

abhi-agg:
> * [x]  Update bergamot translator engine (if required) in the extension
> 
> * [x]  Update the translation worker with the corresponding API changes
> 
> * [x]  Add end to end tests for QE

Fixed by https://github.com/mozilla/firefox-translations/pull/191

PR #191, opened 2022-03-28
Integrate Supervised QE models

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/134

- Updated bergamot translator to [v0.4.3](https://github.com/mozilla/bergamot-translator/releases/tag/v0.4.3)
- Update translation worker JS code to use supervised QE models when available

Tested it via Spanish version of bookings.com website using Outbound Translation as En -> Es has a supervised Quality Model.

The forward translation happens without supervised QE model:

Screenshot 2022-03-28 at 19 09 38

The forward pass of outbound translation happens with supervised Quality Model:

Screenshot 2022-03-28 at 19 10 51


eu9ene (changes requested):
Make sure to report that a supervised model is being used in telemetry. Currently, it is hardcoded in mediator https://github.com/mozilla/firefox-translations/blob/b4609b95a3da290b11f21c94f960d2a4614bb147/extension/mediator.js#L232. You might need to change the message flow if you have this knowledge in the worker only.

andrenatal:
Tests not only neeed to pass, but needs coverage as well

andrenatal (changes requested):
We need tests coverage

abhi-agg:
> Make sure to report that a supervised model is being used in telemetry. Currently, it is hardcoded in mediator
> 
> https://github.com/mozilla/firefox-translations/blob/b4609b95a3da290b11f21c94f960d2a4614bb147/extension/mediator.js#L232
> . You might need to change the message flow if you have this knowledge in the worker only.

@eu9ene I fixed it. Please do refer to the comment: https://github.com/mozilla/firefox-translations/pull/191/files#diff-ce3e24cd36ecb87c17863aa23203ed33457bb09e419fd2dd192ba23576d24f50R234-R238.

extension/mediator.js
```diff
@@ -229,13 +229,24 @@ class Mediator {
...
-                    sentScores: message.payload.sentScores
-                });
+
+                /*
+                 * qe metrics either come from Translation or from InPageTranslation. It is difficult to
+                 * unify them as they are produced for different events (post model construction and
+                 * post translation). Differentiating them here based on the properties to unify the metric
+                 * reporting for qe at one place.
+                 */
+                if (Reflect.apply(Object.prototype.hasOwnProperty, message.payload, ["is_supervised"])) {
```
eu9ene, line 239: you don't need to use the same event handler for metrics from InPageTranslation and Translation. You can just add another event like "reportQeIsSupervised" and handle it there

abhi-agg: Done.

eu9ene (approved):

abhi-agg:
@andrenatal I enabled quality estimation feature in the tests. Please feel free to review 👍🏾

scripts/tests/browser_translation_test.js
```diff
@@ -65,12 +66,23 @@ add_task(async function testTranslationBarDisplayed() {
...
         `Text was correctly translated. (${message})`
       );
 
+      /*
+       * check for the quality scores in translated html content.
+       * The score values might change if different model is used.
+       */
+      is(
+        document.getElementById("translationDiv").innerHTML,
+        "<font x-bergamot-sentence-index=\"0\" x-bergamot-sentence-score=\"-0.304852\"><font x-bergamot-word-index=\"0\" x-bergamot-word-score=\"-0.248207\">Hello</font> <font x-bergamot-word-index=\"1\" x-bergamot-word-score=\"-0.361497\">world.</font></font> <font x-bergamot-sentence-index=\"1\" x-bergamot-sentence-score=\"-0.127826\"><font x-bergamot-word-index=\"0\" x-bergamot-word-score=\"-0.0742788\">That's</font> <font x-bergamot-word-index=\"1\" x-bergamot-word-score=\"-0.213941\">a</font> <font x-bergamot-word-index=\"2\" x-bergamot-word-score=\"-0.204642\">test</font> <font x-bergamot-word-index=\"3\" x-bergamot-word-score=\"-0.0640169\">of</font> <font x-bergamot-word-index=\"4\" x-bergamot-word-score=\"-0.00412017\">translation</font> <font x-bergamot-word-index=\"5\" x-bergamot-word-score=\"-0.205956\">tests.</font></font>",
```
eu9ene, line 82: Those scores will change with every model update or even can be non-deterministic if some probabilistic stuff is used inside, so I would suggest not to rely on them. We can check the fact that scores are present in HTML and maybe that they are in the expected range.

abhi-agg:
@eu9ene I fixed the test case. Please take a look again. The test failure is the non-deterministic failures of the iframe translation that we know of already.

scripts/tests/browser_translation_test.js
```diff
@@ -85,9 +87,69 @@ add_task(async function testTranslationBarDisplayed() {
        );
     }
 
+    // check quality estimation
+    const checkQualityEstimation = async (document, message) => {
+      const translation = document.getElementById("translationDiv").innerText;
+      const translatedHTMLWithQEScores = document.getElementById("translationDiv").innerHTML;
+
+      // just check for the translated content before checking quality estimation
```
eu9ene, line 95: this use case is already covered in another test, why repeat it here?

eu9ene (commented):
We should wait until tests are fixed in main to make sure there are no other regression issues

scripts/tests/browser_translation_test.js
abhi-agg: Before going ahead with checking the quality scores of the translated content, I want to make sure that the translation actually happened. Otherwise, it is not apparent that qe test failures happened because translation failed.

eu9ene: why? you will see that exactly the same "is" assertion failed in `checkTranslation` function

eu9ene (approved):

scripts/tests/browser_translation_test.js
abhi-agg: Ideally, QE tests shouldn't really depend on what happens with `checkTranslation` function and it should remain self-sufficient.

PR #196, opened 2022-03-30
Remove the entire onload event handlers from monitorTabLoad

### LINKED ISSUE #145: Issues with the infobar not being displayed due race condition still persist

andrenatal:
fixes #145 

It turns out that we don't need to listen to all onload events since the content scripts are already injected after onLoad:

see https://github.com/mozilla/firefox-translations/blob/main/extension/manifest.json#L58

and https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extensionTypes/RunAt and https://developer.chrome.com/docs/extensions/mv3/content_scripts/#run_time

That's why the event wasn't triggered all the time.

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -64,58 +64,10 @@ const messageListener = async function(message, sender) {
             break;
         case "monitorTabLoad":
 
-            /*
-             * wait until the page within the tab is loaded, and then return
-             * with the tabId to the caller
-             */
-            listeneronUpdatedLoad = (tabId, changeInfo, tab) => {
-                if (tab.id === sender.tab.id || tab.url === sender.tab.url) {
-                    browser.tabs.onUpdated.removeListener(listeneronUpdatedLoad);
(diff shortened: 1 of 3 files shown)


extension/controller/backgroundScript.js
```diff
@@ -64,58 +64,10 @@ const messageListener = async function(message, sender) {
...
-                            details.tabId ,
-                            { command: "responseMonitorTabLoad", tabId: details.tabId }
-                        );
-                    } ,250);
-                }
-            };
-
-            browser.webNavigation.onCompleted.addListener(webNavigationCompletedLoad);
-            browser.tabs.onUpdated.addListener(listeneronUpdatedLoad);
+            browser.tabs.sendMessage(
```
eu9ene, line 66: do we need this roundtrip to bg script at all?

eu9ene (commented):
it works for me

eu9ene (approved):

PR #202, opened 2022-03-31
Sentry integration

### LINKED ISSUE #70: Integrate Sentry (or something else more appropriate) for error reporting
As @eu9ene and I discussed, we should have better error logging and reporting being submitted to us.

eu9ene:
Sentry wrappings are required to capture errors in async methods and event listeners that are not propagated to the main onerror handler.

fixes #70


extension/manifest.json
```diff
@@ -25,6 +25,7 @@
   "background": {
     "persistent": true,
     "scripts": [
+      "sentry.js",
```
andrenatal, line 28: can we add this file to `model\static\errorReporting`?

extension/manifest.json
```diff
@@ -44,6 +45,7 @@
         "<all_urls>"
       ],
       "js": [
+        "sentry.js",
```
andrenatal, line 48: same as above

extension/settings/prod.js
```diff
@@ -2,5 +2,7 @@
 const settings = {
     logTelemetry: false,
     uploadTelemetry: true,
-    sendDebugPing: false
+    sendDebugPing: false,
+    sentryDsn: "https://9c324392ba52495d8e2de947420950a5@o1069899.ingest.sentry.io/6293991",
```
andrenatal, line 6: just a layman's question: what's the mechanism used by sentry to prevent spam here?

andrenatal (changes requested):
lgtm. let's just address the files' placement

extension/settings/prod.js
eu9ene: https://forum.sentry.io/t/how-does-sentry-prevent-spammers/8188 there is no protection on Sentry side, but I set a limit of 30 events per minute and we can also configure server-side filtering of events if we see something weird there

extension/manifest.json
eu9ene: done

andrenatal (approved):

PR #222, opened 2022-04-06
Auto-translate as browse.

### LINKED ISSUE #17: New translation request required per page
Moving from one page to another - e.g. following a link on a translated page - requires a translation be requested again from the infobar. It's reasonable to assume that if a user translated a page, and a successive page is in the same language or is in the same domain, they'd want to see the next page translated too. Worth adding an "always translate" function, or translate pages following on from previously translated pages ?

andrenatal:
fixes #17 

I implemented always translate the tab though, which I found more useful than always translating a site and a language within that tab.

Blocked by: https://github.com/mozilla-l10n/firefox-translations-l10n/pull/7

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -31,6 +31,7 @@ browser.experiments.translationbar.switchOnPreferences();
 let telemetryByTab = new Map();
 let translationRequestsByTab = new Map();
 let outboundRequestsByTab = new Map();
+const translateAsBrowseMap = new Map();
 
 const init = () => {
   Sentry.wrap(async () => {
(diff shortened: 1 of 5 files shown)


andrenatal:
@eu9ene Maybe we should add telemetry for this action too?

extension/controller/backgroundScript.js
```diff
@@ -306,6 +310,18 @@ const messageListener = function(message, sender) {
...
           break;
+        case "translateAsBrowse":
+
+          /*
+           * we received a request to translate as browse. so this means
+           * we need to record both the tabid, the website and the pagelanguage
+           * so that when there's a navigation in this tab, site and samelanguage
+           * we should automatically start the translation
+           */
+          translateAsBrowseMap.set(message.tabId, {
```
eu9ene, line 321: shouldn't we remove this tabid from the map when the tab is unloaded?

extension/view/js/translation-notification-fxtranslations.js
```diff
@@ -25,13 +25,14 @@ window.MozTranslationNotification = class extends MozElements.Notification {
             <checkbox anonid="qualityestimations-check" label="" style="padding-left:5px" oncommand="this.closest('notification').onQeClick();"/>
           </hbox>
           <vbox class="translating-box" pack="center">
-            <hbox>
+            <hbox class="translate-offer-box" align="center">
               <label value="&translation.translatingContent.label;" style="display:none"/>
-              <label anonid="progress-label" value="" style="padding-left:5px"/>     
+              <label anonid="progress-label" value="" style="padding-left:5px;"/>
+              <button class="notification-button primary" label="" anonid="translateAsBrowse" oncommand="this.closest('notification').translateAsBrowse();"/>
```
eu9ene, line 31: "primary" colour feels like too much when translation is enabled, very distracting. Maybe let's use the grey one like for options and survey?
Screen Shot 2022-04-07 at 11 26 04 AM

extension/controller/backgroundScript.js
andrenatal: done

eu9ene:
> @eu9ene Maybe we should add telemetry for this action too?

yes, we should. It should be an event of click, like for checkboxes and a corresponding boolean metric

andrenatal:
@eu9ene can you add it in a different issue, please?

eu9ene (commented):
lgtm, but we should add telemetry right away

eu9ene:
> @eu9ene can you add it in a different issue, please?

I can of course, but it would be the right thing to accompany any changes with corresponding telemetry fixes.

eu9ene (approved):

Issue #131, opened 2022-03-02
Use GCP instead of Github to fetch engine artifacts for firefox-translations


abhi-agg:

abhi-agg:
@eu9ene Downloading the translation engine over the network and loading it to be able to start using it are 2 different things. Which one are you referring to here?

eu9ene:
I tried it on mac book air m1, and the first time loading took way too long. We load the engine on every page and I didn't see any significant delays after so I assume it's downloading.

abhi-agg:
The model, vocab, shortlist and (possibly) qe model are also downloaded in addition to the engine the first time. It could be anyone of them. Are you saying that extremely slow speeds are happening for engine download while models, vocab and shortlist downloads are way faster?

I am asking because I personally never experienced such delays. I don't think it has much to do with artifacts being downloaded from Github.

This task was created just so that all of the dynamic downloads for extension can happen from the same place i.e. GCP (because models are downloaded from GCP currently) and I am not really expecting any significant improvement on download speeds of the engine after this change. Therefore, the slower speeds that you are reporting is a bit worrisome.

eu9ene:
I think I had this impression because the message was "Loading translation engine.". GCP should provide better SLA than github which is not really intended as storage accessible in real-time.

andrenatal:
I noticed slow download speeds too, and could easily tell that whas the
case because the byte counter was being displayed in the infobar and it was
just slow. We can't also count on GH on the long term, better to have them
on gcp

PR #224, opened 2022-04-06
Fetch bergamot translator engine artifacts from GCP instead of Github

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/131


andrenatal (approved):
not tested, but lgtm if you did

abhi-agg:
I tested them manually. Additionally, e2e tests use the `engineRegistryRootURL` path to download the engine artifacts for testing and these tests are also passing. So we are good 👍🏾

PR #248, opened 2022-04-12
More taskcluster implementations

andrenatal:


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "firefox-translations",
+  "name": "firefox_translations",
   "scripts": {
-    "test": "python scripts/tests/e2e-tests.py",
+    "test": "if [[ -z \"${MOZ_AUTOMATION}\" ]]; then python scripts/tests/e2e-tests.py; fi",
```
abhi-agg, line 4: Just for my information. What is `MOZ_AUTOMATION` used for?

abhi-agg (approved):
lgtm.

package.json
andrenatal: It's an env variable that exists only on taskcluster, and we don't need to run the e2e tests there since we are already running on gh actions whenever we release via https://github.com/mozilla-extensions/firefox-translations

PR #251, opened 2022-04-13
Looks up other languages supported by the browser instead of only the default

andrenatal:
This is such a low hanging fruit that can add a lot of value that we should just do it


abhi-agg (approved):

sylvestre:
Excellent, thanks!
Please let me know when it is ready to be tested!

andrenatal:
@sylvestre you can use the PR build until Nightly gets built tonight: https://github.com/mozilla/firefox-translations/actions/runs/2160317681

sylvestre:
works great, thanks :)

PR #270, opened 2022-04-30
Data collection consent page

### LINKED ISSUE #262: Present a disclosure and consent dialog when collecting analytics
Per the addon ops policy: https://extensionworkshop.com/documentation/publish/add-on-policies/

### LINKED ISSUE #289: isMochitest: beware of relying on this variable
The extension defines a flag (isMochitest) that is supposedly set when a unit test is being run. But that flag can also be set outside of unit tests, simply by visiting one of the URLs that are used in the unit tests.

Currently, there is no issue with that, because the only effect of the flag is that a different model is being loaded.
But if you want to ever tie some more powerful functionality to it, then the check needs to be more accurate.

For now it suffices to put a comment at the implementation that warns about the caveat of using the flag.

https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/mediator.js#L35-L41
https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/translation/Translation.js#L46
https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/translation/translationWorker.js#L710-L714
https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/model/modelRegistry.js#L5-L6

andrenatal:
fixes #262 
fixes #289

depends on #269

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -34,22 +34,26 @@ const scrubSentryEvent = ev => {
   return ev;
 }
 
+const initializeSentry = () => {
+  Sentry.init({
+    dsn: settings.sentryDsn,
+    tracesSampleRate: 1.0,
+    debug: settings.sentryDebug,
+    release: `firefox-translations@${browser.runtime.getManifest().version}`,
+    beforeSend: scrubSentryEvent,
+    integrations(integrations) {
+      // integrations will be all default integrations
+      return integrations.filter(function(integration) {
+        return integration.name !== "Breadcrumbs";
+      });
+    },
+  });
+}
+
 window.addEventListener("load", function () {
   browser.storage.local.get({ errorCollectionConsent: true }).then(item => {
     if (item.errorCollectionConsent) {
-      Sentry.init({
-        dsn: settings.sentryDsn,
-        tracesSampleRate: 1.0,
-        debug: settings.sentryDebug,
-        release: `firefox-translations@${browser.runtime.getManifest().version}`,
-        beforeSend: scrubSentryEvent,
-        integrations(integrations) {
-          // integrations will be all default integrations
-          return integrations.filter(function(integration) {
-            return integration.name !== "Breadcrumbs";
-          });
-        },
-      });
+      initializeSentry();
       console.log("Initializing Sentry");
     }
   });
@@ -339,19 +343,7 @@ const messageListener = function(message, sender) {
           browser.storage.local.set({ errorCollectionConsent: message.consent });
           if (message.consent) {
             if (!Sentry.getCurrentHub().getClient()) {
-              Sentry.init({
-                dsn: settings.sentryDsn,
-                tracesSampleRate: 1.0,
-                debug: settings.sentryDebug,
-                release: `firefox-translations@${browser.runtime.getManifest().version}`,
-                beforeSend: scrubSentryEvent,
-                integrations(integrations) {
-                  // integrations will be all default integrations
-                  return integrations.filter(function(integration) {
-                    return integration.name !== "Breadcrumbs";
-                  });
-                },
-              });
+              initializeSentry();
             } else {
               Sentry.getCurrentHub().getClient()
               .getOptions().enabled = true;
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -35,21 +35,25 @@ const scrubSentryEvent = ev => {
 }
 
 window.addEventListener("load", function () {
-  Sentry.init({
-    dsn: settings.sentryDsn,
-    tracesSampleRate: 1.0,
-    debug: settings.sentryDebug,
-    release: `firefox-translations@${browser.runtime.getManifest().version}`,
-    beforeSend: scrubSentryEvent,
-    integrations(integrations) {
-    // integrations will be all default integrations
-    return integrations.filter(function(integration) {
-      return integration.name !== "Breadcrumbs";
-    });
-  },
+  browser.storage.local.get({ errorCollectionConsent: true }).then(item => {
+    if (item.errorCollectionConsent) {
+      Sentry.init({
+        dsn: settings.sentryDsn,
+        tracesSampleRate: 1.0,
+        debug: settings.sentryDebug,
+        release: `firefox-translations@${browser.runtime.getManifest().version}`,
+        beforeSend: scrubSentryEvent,
+        integrations(integrations) {
+          // integrations will be all default integrations
+          return integrations.filter(function(integration) {
+            return integration.name !== "Breadcrumbs";
+          });
+        },
+      });
+      console.log("Initializing Sentry");
+    }
   });
 });
-
 
 let cachedEnvInfo = null;
 let pingSender = new PingSender();
@@ -61,12 +65,16 @@ let telemetryByTab = new Map();
 let translationRequestsByTab = new Map();
 let outboundRequestsByTab = new Map();
 const translateAsBrowseMap = new Map();
+let isMochitest = false;
 
 const init = () => {
   Sentry.wrap(async () => {
     platformInfo = await browser.runtime.getPlatformInfo();
     cachedEnvInfo = await browser.experiments.telemetryEnvironment.getFxTelemetryMetrics();
     telemetryByTab.forEach(t => t.environment(cachedEnvInfo));
+  });
+  browser.storage.local.get({ telemetryCollectionConsent: true }).then(item => {
+    pingSender.setUploadEnabled(item.telemetryCollectionConsent);
   });
 }
 
@@ -132,6 +140,7 @@ const messageListener = function(message, sender) {
           browser.tabs.sendMessage(sender.tab.id, {
             command: "responseDetectPageLanguage",
             pageLanguage,
+            isMochitest
           })
           break;
         }
@@ -326,6 +335,38 @@ const messageListener = function(message, sender) {
             translatingAsBrowse: message.translatingAsBrowse
           });
           break;
+        case "errorCollectionConsent":
+          browser.storage.local.set({ errorCollectionConsent: message.consent });
+          if (message.consent) {
+            if (!Sentry.getCurrentHub().getClient()) {
+              Sentry.init({
+                dsn: settings.sentryDsn,
+                tracesSampleRate: 1.0,
+                debug: settings.sentryDebug,
+                release: `firefox-translations@${browser.runtime.getManifest().version}`,
+                beforeSend: scrubSentryEvent,
+                integrations(integrations) {
+                  // integrations will be all default integrations
+                  return integrations.filter(function(integration) {
+                    return integration.name !== "Breadcrumbs";
+                  });
+                },
+              });
+            } else {
+              Sentry.getCurrentHub().getClient()
+              .getOptions().enabled = true;
+              console.log("Sentry enabled");
+            }
+          } else {
+            Sentry.getCurrentHub().getClient()
+            .getOptions().enabled = false;
+            console.log("Sentry disabled");
+          }
+          break;
+        case "telemetryCollectionConsent":
+          browser.storage.local.set({ telemetryCollectionConsent: message.consent });
+          pingSender.setUploadEnabled(message.consent);
+          break;
         default:
           // ignore
           break;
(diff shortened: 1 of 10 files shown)


extension/controller/backgroundScript.js
```diff
@@ -33,6 +33,20 @@ let translationRequestsByTab = new Map();
...
+browser.storage.local.get("telemetryCollectionConsent").then(item => {
+  if (Object.keys(item).length !== 0) {
+    settings.uploadTelemetry = item.telemetryCollectionConsent;
+  }
+});
+
+browser.storage.local.get("errorCollectionConsent").then(item => {
+  if (Object.keys(item).length !== 0) {
+    Sentry.getCurrentHub().getClient()
+      .getOptions().enabled = item.errorCollectionConsent;
```
andrenatal, line 46: The Telemetry disable above works, but not this (found this command in a couple of references). `Sentry.init` is called in three classes, so will we need to send three messages?  Maybe there's another way to disable it? @eu9ene

andrenatal:
fixes #262

extension/view/static/dataConsent.html
```diff
@@ -0,0 +1,110 @@
...
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: #ccc;
+  -webkit-transition: .4s;
+  transition: .4s;
+}
+
+.slider:before {
```
cadeyrn, line 33: nit: please use `::before` instead of `:before` because it's a pseudo element and not a pseudo class.

extension/view/static/dataConsent.html
```diff
@@ -0,0 +1,110 @@
...
+/* The slider */
+.slider {
+  position: absolute;
+  cursor: pointer;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: #ccc;
+  -webkit-transition: .4s;
```
cadeyrn, line 29: Do you target such old browsers that a -webkit- variant is needed? According to https://caniuse.com/?search=transition it should be fairly safe to only use the variant without vendor prefix.

extension/controller/backgroundScript.js
```diff
@@ -322,6 +336,15 @@ const messageListener = function(message, sender) {
             translatingAsBrowse: message.translatingAsBrowse
           });
           break;
+        case "errorCollectionConsent":
+          Sentry.getCurrentHub().getClient()
+            .getOptions().enabled = message.consent;
+          browser.storage.local.set({ errorCollectionConsent: message.consent });
+          break;
+        case "telemetryCollectionConsent":
+          settings.uploadTelemetry = message.consent;
```
eu9ene, line 345: PingSender is initialized before this. You should call `pingSender._setUploadEnabled(false)`. There is a possible race condition with loading from prefs. Or maybe we don't need to sync with Firefox prefs anymore and can remove it? https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/model/telemetry/PingSender.js#L48

extension/controller/backgroundScript.js
eu9ene: Sentry is initialized separately in bgScript, contenScript(mediator) and web worker. Since race conditions are possible, we should add message flow to notify mediator and web worker to disable sentry there in the same way.

extension/controller/backgroundScript.js
andrenatal: Let's keep the code to use the prefs setting but replace by the user consent for now in the case things change....again.

extension/view/static/dataConsent.html
andrenatal: fixed, thanks!

andrenatal:
Ok, this is ready to go:

- It fixes the issue that sentry wasn't working in the mediator (https://github.com/mozilla/firefox-translations/issues/305)
- It saves and loads the consent per item from localstorage
- The page is both opened first time the user runs the extension as also displayed in the addon preferences page [1]
- It communicates all three sentry instances and telemetry when the user changes the consent as also loads it first time (all tested)
- Removes the code that controls the telemetry based on Firefox's global telemetry consent
- Incorporated localization - (depends on https://github.com/mozilla-l10n/firefox-translations-l10n/pull/12) 

[1]
Screen Shot 2022-05-03 at 2 45 29 AM

extension/view/static/dataConsent.html
cadeyrn: It seems that you changed one instance of `:before` to `::before`, there are still a few more. :)

extension/controller/backgroundScript.js
```diff
@@ -293,6 +307,25 @@ const messageListener = function(message, sender) {
...
+                command: "errorCollectionConsent",
+                consent: message.consent
+              }
+            );
+          }
+          break;
+        case "telemetryCollectionConsent":
+          browser.storage.local.set({ telemetryCollectionConsent: message.consent });
+          pingSender._setUploadEnabled(message.consent);
+          settings.uploadTelemetry = message.consent;
```
eu9ene, line 327: do not change settings, it blocks sending deletion ping because PingSender thinks it's a debugging build

extension/controller/backgroundScript.js
```diff
@@ -293,6 +307,25 @@ const messageListener = function(message, sender) {
...
+              {
+                command: "errorCollectionConsent",
+                consent: message.consent
+              }
+            );
+          }
+          break;
+        case "telemetryCollectionConsent":
+          browser.storage.local.set({ telemetryCollectionConsent: message.consent });
+          pingSender._setUploadEnabled(message.consent);
```
eu9ene, line 326: remove `_` in method name, since it's a public method now

extension/controller/backgroundScript.js
```diff
@@ -34,6 +40,14 @@ let translationRequestsByTab = new Map();
 let outboundRequestsByTab = new Map();
 const translateAsBrowseMap = new Map();
 
+// set data collection accordingly with what the user set
+browser.storage.local.get("telemetryCollectionConsent").then(item => {
+  if (Object.keys(item).length !== 0) {
+    settings.uploadTelemetry = item.telemetryCollectionConsent;
```
eu9ene, line 46: do not change settings, it blocks sending deletion ping because PingSender thinks it's a debugging build

eu9ene (changes requested):

eu9ene:
disabling telemetry looks working but deletion ping is not sent (see comments). I didn't test disabling of sentry

extension/controller/backgroundScript.js
Rob--W: The `Object.keys(item).length !== 0` approach looks odd. It kind of works because you're looking up one key, so `!== 0` means that the pref was set and the other case would mean that there was nothing saved yet.

Does this logic match with what you have in mind?
It reads prefs and defaults to false (`: false,`) when not stored.

```js
browser.storage.local.get({
  telemetryCollectionConsent: false,
  errorCollectionConsent: false,
}).then(items => {
  if (items.telemetryCollectionConsent) {
    // enable it (assuming that it was disabled by default)
  }
  if (items.errorCollectionConsent) {
    // enable it (assuming that it was disabled by default)
  }
});
```

extension/controller/backgroundScript.js
```diff
@@ -293,6 +307,25 @@ const messageListener = function(message, sender) {
             translatingAsBrowse: message.translatingAsBrowse
           });
           break;
+        case "errorCollectionConsent":
+          browser.storage.local.set({ errorCollectionConsent: message.consent });
+          Sentry.getCurrentHub().getClient()
+            .getOptions().enabled = message.consent;
+          for (const [key,] of telemetryByTab) {
```
Rob--W, line 314: `key` -> `tabId` for clarity.

If telemetry is disabled, shouldn't you just wipe all telemetry?

extension/mediator.js
```diff
@@ -43,6 +32,20 @@ class Mediator {
     }
 
     init() {
+        Sentry.init({
```
Rob--W, line 35: Can't you only init when consent has been granted, so that we know that Sentry is disabled for sure?

Rob--W (commented):
Very quick pass, needs a fuller review later. I am not familiar enough with Sentry's API or the telemetry implementation here to tell whether the toggle is effective, that needs more time.

extension/controller/backgroundScript.js
andrenatal: yep, thanks @Rob--W !

extension/controller/backgroundScript.js
andrenatal: It seems the telemetry api already does that whenever it's switched off. @eu9ene can confirm if that's the case.

extension/controller/backgroundScript.js
andrenatal: fixed

extension/controller/backgroundScript.js
andrenatal: fixed

extension/mediator.js
andrenatal: This will need to change and be refactored based on the outcome of: https://github.com/mozilla/firefox-translations/pull/269

andrenatal:
> Very quick pass, needs a fuller review later. I am not familiar enough with Sentry's API or the telemetry implementation here to tell whether the toggle is effective, that needs more time.

I tested this toggle and it's the official way the recommend to switch it on/off: https://getsentry.github.io/sentry-javascript/interfaces/types.options-1.html#enabled

wagnerand:
cc @kewisch

extension/controller/backgroundScript.js
eu9ene: telemetry sends data deletion ping and disables further uploading when you call `setUplpoadEnabled(false)`

extension/controller/backgroundScript.js
```diff
@@ -326,6 +343,24 @@ const messageListener = function(message, sender) {
...
+              {
+                command: "errorCollectionConsent",
+                consent: message.consent
+              }
+            );
+          }
+          break;
+        case "telemetryCollectionConsent":
+          browser.storage.local.set({ telemetryCollectionConsent: message.consent });
+          pingSender.setUploadEnabled(message.consent);
```
eu9ene, line 362: `_setUploadEnabled` is not renamed to `setUploadEnabled` yet

extension/controller/backgroundScript.js
```diff
@@ -326,6 +343,24 @@ const messageListener = function(message, sender) {
             translatingAsBrowse: message.translatingAsBrowse
           });
           break;
+        case "errorCollectionConsent":
+          browser.storage.local.set({ errorCollectionConsent: message.consent });
+          Sentry.getCurrentHub().getClient()
+            .getOptions().enabled = message.consent;
+          for (const [key,] of telemetryByTab) {
```
eu9ene, line 350: it seems we don't need to send this to tabs anymore, sentry is present only in bgScript

extension/controller/backgroundScript.js
```diff
@@ -61,6 +65,15 @@ let telemetryByTab = new Map();
 let translationRequestsByTab = new Map();
 let outboundRequestsByTab = new Map();
 const translateAsBrowseMap = new Map();
+let isMochitest = false;
+
+// set data collection accordingly with what the user set
+browser.storage.local.get({ telemetryCollectionConsent: true,
+  errorCollectionConsent: true, }).then(items => {
+    pingSender.setUploadEnabled(items.telemetryCollectionConsent);
+    Sentry.getCurrentHub().getClient()
```
eu9ene, line 74: hm, we're already doing this in window.load after sentry init... not sure what will happen if it's not initialized yet

extension/controller/backgroundScript.js
```diff
@@ -326,6 +335,38 @@ const messageListener = function(message, sender) {
             translatingAsBrowse: message.translatingAsBrowse
           });
           break;
+        case "errorCollectionConsent":
+          browser.storage.local.set({ errorCollectionConsent: message.consent });
+          if (message.consent) {
+            if (!Sentry.getCurrentHub().getClient()) {
+              Sentry.init({
```
eu9ene, line 342: I suggest extracting function with Sentry init

eu9ene, line 342: what will happen if we init it the second time?

eu9ene (dismissed):
Overall looks good, but please make sure we don't reinitialize sentry

extension/controller/backgroundScript.js
andrenatal: done

andrenatal: That won't happen. `if (!Sentry.getCurrentHub().getClient()) ` will safegaurd it

eu9ene (approved):

Issue #267, opened 2022-04-29
Replacement for unsafe scratch.innerHTML


Rob--W:
As mentioned during today's review meeting, there is a highly unsafe use of innerHTML that enables script execution at [extension/view/js/InPageTranslation.js#L535-L536](https://github.com/mozilla/firefox-translations/blob/e4fd84ec938df868143aeb5b9161a704d61bd79c/extension/view/js/InPageTranslation.js#L535-L536).

```js
 const scratch = node.cloneNode(false); // shallow clone of parent node 
 scratch.innerHTML = translatedHTML; 
```

To make this safer, the following replacement can be used. There are some caveats, see below.

```js
const nonLiveDomContainer = document.createElement("template");
nonLiveDomContainer.innerHTML = translatedHTML;
const scratch = nonLiveDomContainer.content;
```

In the original code, `scratch` is an `Element` instance, in the new one it's a `DocumentFragment`. They have different interfaces, but the most important one, [`src.childNodes`](https://github.com/mozilla/firefox-translations/blob/e4fd84ec938df868143aeb5b9161a704d61bd79c/extension/view/js/InPageTranslation.js#L580) is supported. It looks like you're [using `src.outerHTML` for debugging](https://github.com/mozilla/firefox-translations/blob/e4fd84ec938df868143aeb5b9161a704d61bd79c/extension/view/js/InPageTranslation.js#L652) which is not available on the DocumentFragment. For debugging, `src.outerHTML ?? src` can be used instead.

I believe that this replacement should be good enough for you, but if you want `scratch` to be an exact copy of the element, then you could concatenate `node.cloneNode(false)`'s start tag with `translatedHTML` before assignment to `nonLiveDomContainer.innerHTML`, and get that node with `nonLiveDomContainer.content.firstChild`.

While the change above fixes the immediate issue with that specific line, an unsafe situation can happen again whenever any part of `scratch` is turned into live DOM, which happens when elements within are appended to the live document. This happens and needs to be resolved (separately) at https://github.com/mozilla/firefox-translations/blob/e4fd84ec938df868143aeb5b9161a704d61bd79c/extension/view/js/InPageTranslation.js#L599

Relevant docs:
- https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

Rob--W:
> In none of those I was able to escalate from the page execution context into the extension execution context.

[...]

PR #281, opened 2022-05-02
Replaces unsafe use of innerHTML and appendChild

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/267

 - Sanitize the qe specific html text that is added to DOM via `appendChild`
 - Using DocumentFragment to update DOM instead of setting innerHTML directly
 
 cc @Rob--W

### DIFF
--- a/extension/view/js/InPageTranslation.js
+++ b/extension/view/js/InPageTranslation.js
@@ -27,6 +27,10 @@ class InPageTranslation {
         this.withOutboundTranslation = null;
         this.withQualityEstimation = null;
         this.QE_THRESHOLD = Math.log(0.5);
+        this.qeAttributes = new Set([
+            "x-bergamot-sentence-index", "x-bergamot-sentence-score",
+            "x-bergamot-word-index", "x-bergamot-word-score",
+        ]);
 
         /*
          * reference for all tags:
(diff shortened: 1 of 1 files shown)


extension/view/js/InPageTranslation.js
```diff
@@ -589,13 +594,29 @@ class InPageTranslation {
...
+                                    "x-bergamot-sentence-index", "x-bergamot-sentence-score",
+                                    "x-bergamot-word-index", "x-bergamot-word-score",
+                                ]);
+
+                                if (node.nodeType === Node.ELEMENT_NODE) {
+                                    if (node.nodeName !== "FONT") return false;
+
+                                    if (!node.getAttributeNames().every(attribute => allowed.has(attribute))) return false;
+
+                                    for (let child of node.childNodes) if (!hasOnlyQEAttributes(child)) return false;
```
Rob--W, line 608: Iterate over `.children` instead of `.childNodes`.

`.children` contains elements only, `childNodes` also contains non-elements.

extension/view/js/InPageTranslation.js
```diff
@@ -589,13 +594,29 @@ class InPageTranslation {
...
                         if (!counterpart) {
 
+                            const hasOnlyQEAttributes = node => {
+                                const allowed = new Set([
+                                    "x-bergamot-sentence-index", "x-bergamot-sentence-score",
+                                    "x-bergamot-word-index", "x-bergamot-word-score",
+                                ]);
+
+                                if (node.nodeType === Node.ELEMENT_NODE) {
+                                    if (node.nodeName !== "FONT") return false;
```
Rob--W, line 604: `.nodeName` is upper-case for HTML, but in XHTML documents it's case-sensitive. You'd either need to compare case-insensitively, or make sure that the generated HTML is all uppercase `<FONT>` instead of `<font>`.

Rob--W (changes requested):

abhi-agg:
@Rob--W I made changes as per your comments. Please feel free to review it again. I hope I didn't miss anything 👍🏾

PR #293, opened 2022-05-02
Register translation-notification CE once per window

### LINKED ISSUE #273: Improving custom element initialization in TranslationBar/api.js / translation-notification-fxtranslations.js
The notification bar is implemented with a custom element, but in practice the registered element is used only once (and registered repeatedly). The `chromeWin.now` variable is added and used, allegedly "to load updates for the translation notification on extension reload" ([source in TranslationBar/api.js](https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/experiments/TranslationBar/api.js#L77-L100)). Since custom elements can be registered only once per window, and extensions may update, there is indeed a need for a non-fixed name, but it suffices to generate that name just once instead of anew for every custom element.

To implement the above suggestions, modify extension/controller/experiments/TranslationBar/api.js as follows:

- Generate a unique name at the top of the file, e.g.
  ```js
  const translationElementId = `translation-notification-${Date.now()}`; 
   ```
- Declare a `WeakSet` to hold `chromeWin` references that have already got a custom element registration.
  - This must be at the top of TranslationBar/api.js, and not inside the `getAPI` method like your current implementation, because the set should be globally unique in order to guarantee that each window gets only one custom element registration by this version of the extension.
- In `translationbar`'s `show` method implementation, check if the `chromeWin` is already part of the set. If it is not, load the custom element definition in the window and `chromeWin` to the set.
  - I wouldn't use `chromeWin.customElements.setElementCreationCallback`, but directly register the custom element since it seems that you'll immediately be using the custom element, and there is no apparent gain from registering the element lazily.
  - Avoid the use of `window.now`, because it is a too generic name that can easily cause surprising conflicts later on.
     - An easy way is to rename `chromeWin.now` to something more specific, e.g. `translationElementIdSuffix`.
     - An alternative is to load translation-notification-fxtranslations.js in TranslationBar/api.js, and let it export a function that takes the `chromeWin` object, and make sure that any references to the globals within the window are fixed:
        - ```js
          function defineTranslationElement(window, translationElementIdSuffix) {
            const { MozElements, document, customElements, Services } = window;
            // Rest of translation-notification-fxtranslations.js here, with window.now replaced by translationElementIdSuffix
          }

Rob--W:
Fixes #273

### DIFF
--- a/extension/controller/experiments/TranslationBar/api.js
+++ b/extension/controller/experiments/TranslationBar/api.js
@@ -7,6 +7,15 @@
 
 /* global ExtensionAPI, ChromeUtils, modelRegistry, TranslationNotificationManager  */
 
+/*
+ * custom elements can only be registered, and not unregistered.
+ * To make sure that the extension is able to register the custom element
+ * even after an extension update/reload, use a generated unique name.
+ */
+const TRANSLATION_NOTIFICATION_ELEMENT_ID = `translation-notification-${Date.now()}`;
+const windowsWithCustomElement = new WeakSet();
+
+
  // eslint-disable-next-line no-invalid-this
  this.experiments_translationbar = class extends ExtensionAPI {
     getAPI(context) {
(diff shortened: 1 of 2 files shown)


Rob--W:
I have verified that there are no regressions and that the element is registered once as desired:

1. Load extension, view translatebar - it works.
2. Run `window.TRANSLATION_NOTIFICATION_ELEMENT_ID` from browser console; it prints the expected name.
3. Reload the extension, open new tab, view translatebar - it still works.
4. Run `window.TRANSLATION_NOTIFICATION_ELEMENT_ID` from the browser console, it prints a new ID.
5. Open a new tab, view translatebar - it works too.
6. Run `window.TRANSLATION_NOTIFICATION_ELEMENT_ID` from the browser console, it prints the same ID.

andrenatal (approved):

PR #297, opened 2022-05-02
Disable translation cache

### LINKED ISSUE #288: Ensure that translations are independent when cross-origin/cross-frame data is involved
To translate, the current implementation aggregates the source text from all frames from a tab, forwards it to the top-level frame, where a worker is spawned that does the translation, and then the messages are returned to the original frames. Even if #287 is fixed, there is still a possibility of cross-origin data leakage, when the translations are not independent. That is, if the translation engine maintains any state between translations, with the latest translation being informed/derived from data from previous translations, then a cross-origin info leak may occur.

@andrenatal mentioned that there is (or used to be) a cache that supports the translation. If this is indeed the case, the cache should be wiped as soon as the `frameId` of the source text to translate changes.

In general, any origin-specific state should be cleared when the worker processes a translation request from a different frame than before.

eu9ene:
fixes #288

### DIFF
--- a/extension/controller/translation/translationWorker.js
+++ b/extension/controller/translation/translationWorker.js
@@ -302,7 +302,8 @@ class TranslationHelper {
         // instantiate the Translation Service
         constructTranslationService() {
             if (!this.translationService) {
-                let translationServiceConfig = { cacheSize: 10 };
+                // caching is disabled (see https://github.com/mozilla/firefox-translations/issues/288)
+                let translationServiceConfig = { cacheSize: 0 };
                 console.log(`Creating Translation Service with config: ${translationServiceConfig}`);
                 this.translationService = new this.WasmEngineModule.BlockingService(translationServiceConfig);
                 console.log("Translation Service created successfully");


extension/controller/translation/translationWorker.js
```diff
@@ -302,7 +302,8 @@ class TranslationHelper {
         // instantiate the Translation Service
         constructTranslationService() {
             if (!this.translationService) {
-                let translationServiceConfig = { cacheSize: 10 };
+                // caching is disabled (see https://github.com/mozilla/firefox-translations/issues/288)
+                let translationServiceConfig = { cacheSize: 0 };
```
Rob--W, line 306: I was wondering whether `0` would result in a fallback to a non-zero value, which is fortunately not the case:
https://github.com/mozilla/bergamot-translator/blob/5ae1b1ebb3fa9a3eabed8a64ca6798154bd486eb/src/translator/service.cpp#L34-L43

hence r+

Rob--W (approved):

andrenatal (approved):

PR #299, opened 2022-05-03
Add check of frame origin on recieving translation results

### LINKED ISSUE #287: translate / translationComplete messages are sent across frames, potentially to different documents
- ["translate" is sent to the top frame](https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/backgroundScript.js#L183-L191). Presumably the handler/worker translates the input and routes it back to the sender.
- Upon completion, ["translationComplete" is reported back to the original frame](https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/backgroundScript.js#L197-L204).

There is no verification that the original frame still contains the original document, which can result in information being leaked across origins.

At minimum, a patch for this bug would be to include [`window.origin`](https://developer.mozilla.org/en-US/docs/Web/API/origin) in the original message, and to verify that it's the same before processing the response when received later, in `mediator.js`, in every kind of message where the `frameId` is included.

`window.origin` is `"null"` when the document is sandboxed. To avoid leakage between different domains when sandboxed, I suggest to fall back to `location.origin` when `window.origin` is `"null"`. Note that we should not unconditionally use `location.origin` without checking `window.origin`, because the former would report the wrong results for `about:srcdoc` and `about:blank` URLs, which can contain data from different origins.

eu9ene:
fixes #287 

@Rob--W please take a look, I hope I got it right. 

This should prevent leakage of information from translation messages if a frame has changed its origin. Other messages sent to specific frames are broadcasting of general info. They are delayed for some frames because of their loading, so handling `translationComplete` should be enough.


extension/mediator.js
```diff
@@ -19,6 +19,11 @@ window.addEventListener("load", function () {
   });
 });
 
+const getOrigin = () => {
+    return window.origin !== null
```
Rob--W, line 23:
```suggestion
    return window.origin !== "null"
```

Here is a test page; it has the `Content-Security-Policy: sandbox allow-scripts` header: https://robwu.nl/cors-anywhere.html

You can also test this even if the page doesn't set the header by specifying `<iframe src="https://example.com/" sandbox=""></iframe>`, as long as the sandbox attribute does not contain "allow-same-origin".

For documentation on the sandbox property, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox

P.S. Since the value of this is a constant, you may as well change this to `const THIS_ORIGIN = window.origin != "null" ? window.origin : location.origin;`

extension/mediator.js
```diff
@@ -338,7 +345,12 @@ class Mediator {
                     this.translate(message)
                     break
                 case "translationComplete":
-                    this.updateElements(message.translationMessage);
+                    if (getOrigin() === message.translationMessage.origin) {
+                        this.updateElements(message.translationMessage);
+                    } else {
+                        console.warn(`Message with a different origin is recieved by a frame.
```
Rob--W, line 351: Just for clarity: this issue is not specific to frames. A top-level navigation can also have the same issue, because tabId doesn't change and frameId will still be `0`.

Rob--W (changes requested):
The `!== null` check should be `!== "null"`.

extension/mediator.js
eu9ene: fixed

extension/mediator.js
eu9ene: I added an origin check before updating HTML content for all "translationComplete" messages, not only frame ones

extension/mediator.js
```diff
@@ -140,13 +143,15 @@ class Mediator {
         switch (message.command) {
             case "translate":
                 if (this.isMainFrame) {
+                    if (!("origin" in message)) message.origin = THIS_ORIGIN;
```
Rob--W, line 146: Is it possible to make this unconditional, by putting it in the right place where the origin of a message is unambiguous?

eu9ene: Actually, it shouldn't be conditional, fixed. I think it's the right place, otherwise, we'll need to add this to both InPageTranslation and OutboundTranslation. I prefer to keep this logic in mediator.

Rob--W (approved):
Current version of the code has proper checks.

PR #300, opened 2022-05-03
Remove tooltip from low quality translated elements (QE)

### LINKED ISSUE #290: All tooltips have been replaced with "Sentence: <score>  Word: <score>"
When "Enable quality estimation" is enabled, all tooltips on the web page are replaced with `Sentence: <score>  Word: <score>`. This may reduce the usability of a web page.

If it is not possible to avoid the assignment of the new title, then it may make sense to at least concatenate the existing title (if any) with the new one (separated by a `"\n"` if there was already a title).

https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/view/js/InPageTranslation.js#L221-L225

andrenatal:
fixes #290 

In this current form this presents no value to the user.

### DIFF
--- a/extension/view/js/InPageTranslation.js
+++ b/extension/view/js/InPageTranslation.js
@@ -217,13 +217,6 @@ class InPageTranslation {
                 el.classList.toggle("bad",true);
             }
         });
-
-        // add tooltips to each (sub)word with sentence and word score.
-        document.querySelectorAll("[x-bergamot-sentence-score] > [x-bergamot-word-score]").forEach(el => {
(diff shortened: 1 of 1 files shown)


Rob--W (approved):
r+ because this solves the issue, but I don't have enough context to judge whether this is desirable from the functionality's POV, and will defer to other reviewers for that aspect.

andrenatal:
Yes, we don't need those tooltips. The end user will have no idea what that means.

PR #302, opened 2022-05-03
Also utilize getComputedStyle() to determine if an element is hidden

### LINKED ISSUE #279: isElementHidden only looks at inline styles and is ignorant about stylesheets
`isElementHidden` is not effective/accurate - it only examines some inline style attributes. If truly interested in the styles, use `window.getComputedStyle(element)` instead of `element.style`.

https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/view/js/InPageTranslation.js#L292-L293

andrenatal:
fixes #279

### DIFF
--- a/extension/view/js/InPageTranslation.js
+++ b/extension/view/js/InPageTranslation.js
@@ -290,8 +290,9 @@ class InPageTranslation {
     }
 
     isElementHidden(element) {
-        return window.getComputedStyle(element).display === "none" ||
-                window.getComputedStyle(element).visibility === "hidden" ||
+        const computedStyle = window.getComputedStyle(element);
+        return computedStyle.display === "none" ||
+                computedStyle.visibility === "hidden" ||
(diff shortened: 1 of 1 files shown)


extension/view/js/InPageTranslation.js
```diff
@@ -290,7 +290,9 @@ class InPageTranslation {
     }
 
     isElementHidden(element) {
-        return element.style.display === "none" || element.style.visibility === "hidden" || element.offsetParent === null;
+        return window.getComputedStyle(element).display === "none" ||
```
Rob--W, line 293: Let's call `getComputedStyle` once instead of twice.

```js
let style = window.getComputedStyle(element);
return style.display === "none" || style.visibility = etc.
```

Rob--W (changes requested):

Rob--W (approved):

PR #304, opened 2022-05-03
Use DOMPurify to escape outbound translations widget labels

### LINKED ISSUE #280: Replace innerHTML usage in view/js/OutboundTranslation.js
The formtranslationsDescription string may contain HTML such as `<em>`, e.g. from https://github.com/mozilla-l10n/firefox-translations-l10n/blob/86c4cd84479bacda6b126a682b361059b1eba443/en_US/messages.json#L103. Even if we assume the translation to be trusted, `pageLanguage`/`navigatorLanguage` vars are potentially external and should be HTML-escaped to avoid risk.

https://github.com/mozilla/firefox-translations/blob/e4fd84ec938df868143aeb5b9161a704d61bd79c/extension/view/js/OutboundTranslation.js#L44-L45

The following `innerHTML` usage can be replaced with `.textContent`. I looked up all callers, the strings are from firefox-translations-l10n repo and are strings without HTML.

https://github.com/mozilla/firefox-translations/blob/e4fd84ec938df868143aeb5b9161a704d61bd79c/extension/view/js/OutboundTranslation.js#L291-L293

andrenatal:
fixes #280


Rob--W (changes requested):
Looks like you've pulled the copy of DOMPurify from the main branch of the repo.

Use an official release instead: https://github.com/cure53/DOMPurify/releases/tag/2.3.6

Otherwise the file will not be flagged as a known library by the linter, which complicates the add-on review process.

Rob--W (approved):

PR #307, opened 2022-05-03
Clarify shared state of LanguageDetection

### LINKED ISSUE #284: languageDetection sends unnecessarily much data across processes
The implementation of language detection instantiates a `LanguageDetection` class, and repeatedly sends the (implicit serialization of the class, i.e. its properties) to the background script (backgroundScript.js) and back to the content script (mediator.js). Since the initial `LanguageDetection` instance contains the page's full text, this means that a potentially large chunk of data is repeatedly send over IPC, which is inefficient. In practice, it looks like the body is only needed once, for the initial language detection, and after that it's not needed any more, so the implementation can easily be improved without loss of functionality by only passing the properties that it needs.

The `responseDetectPageLanguage` message from the second block below does not specify a `frameId`, so that message is broadcast to all frames in the tab, which worsens the impact of this bug.

The code relevant to this issue is:

https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/mediator.js#L56-L65

https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/backgroundScript.js#L76-L108

https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/mediator.js#L316-L318

### LINKED ISSUE #296: pageAction.onClicked may not work, or use the language detection results from the wrong tab
The `pageAction.onClicked` handler relies on a global `languageDetection` variable.
That logic is very flawed and has two issues:
- It does not work when the extension has just been installed.
- A tab load in the background can change the global. Then, upon clicking the pageAction button the translation will use the incorrect context.

https://github.com/mozilla/firefox-translations/blob/7dcdb50ba4e906f4aaee0bfe19d2fc0e668b7018/extension/controller/backgroundScript.js#L357-L366

via the following (call stack)

https://github.com/mozilla/firefox-translations/blob/7dcdb50ba4e906f4aaee0bfe19d2fc0e668b7018/extension/controller/backgroundScript.js#L75-L79
https://github.com/mozilla/firefox-translations/blob/7dcdb50ba4e906f4aaee0bfe19d2fc0e668b7018/extension/mediator.js#L50-L66
https://github.com/mozilla/firefox-translations/blob/7dcdb50ba4e906f4aaee0bfe19d2fc0e668b7018/extension/mediator.js#L313-L314
https://github.com/mozilla/firefox-translations/blob/7dcdb50ba4e906f4aaee0bfe19d2fc0e668b7018/extension/controller/backgroundScript.js#L110-L116
https://github.com/mozilla/firefox-translations/blob/7dcdb50ba4e906f4aaee0bfe19d2fc0e668b7018/extension/mediator.js#L46

Rob--W:
- Remove global `LanguageDetection` instance from backgroundScript.js, so that it is obvious that the `pageAction.onClicked` listener does not depend on any tab-specific state (#296).

- Minimize the amount of data (i.e. wordsToDetect) that is being passed back and forth, to only when needed (#284).

Fixes #284
Fixes #296

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -24,7 +24,6 @@ let cachedEnvInfo = null;
 let pingSender = new PingSender();
 let modelFastText = null;
 let modelFastTextReadyPromise = null;
-let languageDetection = null;
 let platformInfo = null;
 
 // as soon we load, we should turn off the legacy prefs to avoid UI conflicts
@@ -74,16 +73,14 @@ const messageListener = function(message, sender) {
   // eslint-disable-next-line complexity,max-lines-per-function
     Sentry.wrap(async() => {
       switch (message.command) {
-        case "detectPageLanguage":
+        case "detectPageLanguage": {
           await modelFastTextReadyPromise;
 
-          languageDetection = Object.assign(new LanguageDetection(), message.languageDetection);
-
           /*
-           * if we don't support this browser's language, we nede to hide the
+           * if we don't support this browser's language, we need to hide the
            * page action and bail right away
            */
-          if (!languageDetection.isBrowserSupported()) {
+          if (!message.languageDetection.supported) {
             browser.pageAction.hide(sender.tab.id);
             break;
           }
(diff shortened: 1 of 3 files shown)


andrenatal:
@Rob--W  is this ready to review?

Rob--W:
> @Rob--W  is this ready to review? 

Yes

andrenatal (approved):

PR #308, opened 2022-05-03
Restore original prefs on shutdown

### LINKED ISSUE #277: Restore original prefs upon shutdown
The implementation sets a bunch of preferences when the background starts up:
https://github.com/mozilla/firefox-translations/blob/7230c7f65c3b367d087b31b621bf5cfcbddd22da/extension/controller/experiments/TranslationBar/api.js#L133-L136

Non-extension-specific prefs should be reverted to their original value when the extension is shut down.
To do so, add an `onShutdown` method to the `ExtensionAPI` class. This will be called when the extension shuts down.
It won't be called when the browser crashes, but that's a harder problem to solve.

This logic of storing the original pref in memory and restoring the original pref value on shutdown only works when the extension is the only one in control of the pref. For translation-specific prefs, that seems likely.

`javascript.options.wasm_simd_wormhole` may become redundant if #255 implies the discontinuation of the feature.

Rob--W:
Fixes #277

### DIFF
--- a/extension/controller/experiments/TranslationBar/api.js
+++ b/extension/controller/experiments/TranslationBar/api.js
@@ -5,7 +5,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-/* global ExtensionAPI, ChromeUtils, modelRegistry, TranslationNotificationManager  */
+/* global ExtensionCommon, ExtensionAPI, ChromeUtils, Services, modelRegistry, TranslationNotificationManager  */
+
+// the Services object was not available until Firefox 88 - bugzil.la/1698158
+if (typeof Services === "undefined") {
+  // eslint-disable-next-line no-invalid-this
+  this.Services = ChromeUtils.import("resource://gre/modules/Services.jsm").Services;
+}
 
 /*
  * custom elements can only be registered, and not unregistered.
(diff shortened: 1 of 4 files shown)


extension/controller/experiments/TranslationBar/api.js
```diff
@@ -38,10 +75,6 @@ const translationNotificationManagers = new Map();
       const { ExtensionUtils } = ChromeUtils.import("resource://gre/modules/ExtensionUtils.jsm");
       const { ExtensionError } = ExtensionUtils;
 
-      const { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
-
-      const { ExtensionCommon } = ChromeUtils.import("resource://gre/modules/ExtensionCommon.jsm");
```
Rob--W, line 43: I've removed ExtensionCommon because it is unnecessary due to it being part of the environment by default: https://searchfox.org/mozilla-central/rev/ea1234192518e01694a88eac8ff090e4cadf5ca4/toolkit/components/extensions/ExtensionCommon.jsm#1809,1818

extension/controller/experiments/TranslationBar/api.js
```diff
@@ -5,7 +5,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-/* global ExtensionAPI, ChromeUtils, modelRegistry, TranslationNotificationManager  */
+/* global ExtensionCommon, ExtensionAPI, ChromeUtils, Services, modelRegistry, TranslationNotificationManager  */
+
+// the Services object was not available until Firefox 88 - bugzil.la/1698158
```
Rob--W, line 10: According to manifest.json, the supported min version is Firefox 79, so I have included this for compatibility with Firefox 87 and earlier. If such (1+ year) old Firefox versions are not supported, then this can be removed.

andrenatal:
@Rob--W  is this ready to review?

Rob--W:
> @Rob--W  is this ready to review?

Yes

andrenatal (approved):

Issue #324, opened 2022-05-09
“We found an error loading the translation engine” when attempting to translate a page in Private Browsing


acornestean:
Description:
Attempting to translate a page in Private Browsing fails and the infobar displays the “ We found an error loading the translation engine” message.

An error message is also displayed in the web console: 

`Model 'esen' construction failed: 'The operation is insecure. - getItemFromCacheOrWeb@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:460:40
downloadFiles@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:438:39
constructTranslationModelHelper/<@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:382:89
constructTranslationModelHelper@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:382:18
constructTranslationModel@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:345:28
loadLanguageModel@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:273:26
loadTranslationEngine/initialModule.onRuntimeInitialized<@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/translationWorker.js:79:26
doRun@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/bergamot-translator-worker.js:2882:50
run@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/bergamot-translator-worker.js:2894:7
runCaller@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/bergamot-translator-worker.js:2863:22
removeRunDependency@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/bergamot-translator-worker.js:429:8
receiveInstance@moz-extension://7f6c01e5-da66-4b70-84ec-0f8aecf59c76/controller/translation/bergamot-translator-worker.js:514:26
[...]

Issue #283, opened 2022-05-02
Page action is available on HTTP pages


wagnerand:
The page action is available for HTTP pages, but when you try to apply a translation, you get a note that the add-on only works on HTTPS pages.

Rob--W:
May be preferable to resolve the issues that prevent it from working on HTTP.

andrenatal:
I'm seeing many `self.crypto.randomUUID` on sentry, so yes, let's try to fix this...

abhi-agg:
@wagnerand Could you please share the steps to reproduce it?

lonnen:
To be clear, this issue is about this button: <img width="90" alt="Screen Shot 2022-05-19 at 12 16 58 PM" src="https://user-images.githubusercontent.com/21467/169385512-0eef5e15-a6b4-4e89-94b7-8a6734c21434.png">

Which is controlled by the `browser.pageAction` API

That said, I agree with Rob--W and understand the plan here is to resolve the underlying HTTP issue

andrenatal:
I have just tested https://github.com/mozilla/firefox-translations/pull/330 on http pages (http://www.columbia.edu/~fdc/sample.html, http://www.csun.edu/~sk36711/WWW/tutorials/how_web_pages_work.html and http://info.cern.ch/ for the record) and it just translated fine. So when we sort the problem we are facing there [1], this issue should also be closed. cc @abhi-agg  

[1] https://github.com/mozilla/firefox-translations/pull/330#issuecomment-1132390299

PR #330, opened 2022-05-13
Enable translation in private browsing mode

abhi-agg:
Fixes https://github.com/mozilla/firefox-translations/issues/324 and https://github.com/mozilla/firefox-translations/issues/283

Disabled using cache while browsing in private mode as per https://github.com/mozilla/firefox-translations/issues/324#issuecomment-1123734734

Fixes https://github.com/mozilla/Foxfooding_Firefox_translation_community/issues/7

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -330,11 +330,9 @@ const messageListener = function(message, sender) {
             );
           } catch (error) {
             sendUpdateProgress(message.tabId, ["updateProgress", "errorLoadingWasm"]);
-            //postMessage(["reportError", "model_load"]);
             getTelemetry(message.tabId).record("counter", "errors", "model_load");
             console.warn("Reporting content script error to Sentry");
             Sentry.captureException(deserializeError(serializeError(error)));
-            //postMessage(["reportException", serializeError(error)]);
           }
           break;
         case "updateProgress":
@@ -496,8 +494,12 @@ Promise.allSettled([displayedConsentPromise, isMochitestPromise]).then(values =>
 const getLanguageModels = async (tabId, languagePairs) => {
   console.log(`LangPairs: ${JSON.stringify(languagePairs)}\n modelRegistryRootURL:${modelRegistryRootURL}`);
   let languageModelsPromise = [];
+  let start = performance.now();
   languagePairs.forEach(languagePair => languageModelsPromise.push(getLanguageModel(tabId, languagePair)));
   let languageModelURLs = await Promise.all(languageModelsPromise);
+  let end = performance.now();
+  console.log(`Total Download time for all language model files: ${(end - start) / 1000}s`);
+  getTelemetry(tabId).record("timespan", "performance", "model_download_time_num", end-start);
 
   let result = [];
   languageModelURLs.forEach((languageModelURL, index) => {
@@ -510,7 +512,7 @@ const getLanguageModels = async (tabId, languagePairs) => {
 };
 
 const getLanguageModel = async (tabId, languagePair) => {
-  //console.log(`LangPairs: ${JSON.stringify(languagePairs)}\n modelRegistryRootURL:${modelRegistryRootURL}`);
+  console.log(`LangPair: ${JSON.stringify(languagePair)}`);
   let languageModelPromise = [];
   languageModelFileTypes
       .filter(fileType => fileType !== "qualityModel" || languagePair.withQualityEstimation)
@@ -535,7 +537,6 @@ const downloadFile = async (tabId, fileType, languagePairName) => {
   const buffer = await getItemFromCacheOrWeb(tabId, fileName, fileSize, fileChecksum);
   if (!buffer) {
       console.error(`Error loading models from cache or web ("${fileType}")`);
-      //postMessage(["onError", "model_download"]);
       throw new Error(`Error loading models from cache or web ("${fileType}")`);
   }
   return buffer;
@@ -593,20 +594,12 @@ const getItemFromWeb = async (tabId, itemURL, fileSize, fileChecksum) => {
       console.log(`Error downloading ${itemURL} (error: ${error})`);
       // inform mediator
       sendUpdateProgress(tabId, ["updateProgress", "notfoundErrorsDownloadingEngine"]);
-      /*postMessage([
-          "updateProgress",
-          "notfoundErrorsDownloadingEngine"
-      ]);*/
       return null;
   }
 
   if (!fetchResponse.ok) {
       console.log(`Error downloading ${itemURL} (response status:${fetchResponse.status})`);
       sendUpdateProgress(tabId, ["updateProgress", "notfoundErrorsDownloadingEngine"]);
-      /*postMessage([
-          "updateProgress",
-          "notfoundErrorsDownloadingEngine"
-      ]);*/
       return null;
   }
 
@@ -625,10 +618,6 @@ const getItemFromWeb = async (tabId, itemURL, fileSize, fileChecksum) => {
           if (elapsedTime > MAX_DOWNLOAD_TIME) {
               console.log(`Max time (${MAX_DOWNLOAD_TIME}ms) reached while downloading ${itemURL}`);
               sendUpdateProgress(tabId, ["updateProgress", "timeoutDownloadingEngine"]);
-              /*postMessage([
-                  "updateProgress",
-                  "timeoutDownloadingEngine"
-              ]);*/
               return false;
           }
           // eslint-disable-next-line no-await-in-loop
@@ -644,17 +633,9 @@ const getItemFromWeb = async (tabId, itemURL, fileSize, fileChecksum) => {
               chunks.push(value);
               receivedLength += value.length;
               sendUpdateProgress(tabId, ["updateProgress", ["downloadProgress", [`${receivedLength}`,`${contentLength}`]]]);
-              /*postMessage([
-                  "updateProgress",
-                  ["downloadProgress", [`${receivedLength}`,`${contentLength}`]]
-              ]);*/
           } else {
             sendUpdateProgress(tabId, ["updateProgress", "nodataDownloadingEngine"]);
-              /*postMessage([
-                  "updateProgress",
-                  "nodataDownloadingEngine"
-              ]);*/
-              return false;
+            return false;
           }
 
           if (receivedLength === contentLength) {
@@ -675,20 +656,12 @@ const getItemFromWeb = async (tabId, itemURL, fileSize, fileChecksum) => {
       if (!sha256) {
           console.log(`Sha256 error for ${itemURL}`);
           sendUpdateProgress(tabId, ["updateProgress", "tlsIncompatibility"]);
-          /*postMessage([
-              "updateProgress",
-              "tlsIncompatibility"
-          ]);*/
           return false;
       }
 
       if (sha256 !== fileChecksum) {
           console.log(`Checksum failed for ${itemURL}`);
           sendUpdateProgress(tabId, ["updateProgress", "checksumErrorsDownloadingEngine"]);
-          /*postMessage([
-              "updateProgress",
-              "checksumErrorsDownloadingEngine"
-          ]);*/
           return false;
       }
       console.log(`Checksum passed for ${itemURL}`);
@@ -713,7 +686,6 @@ const digestSha256 = async (buffer) => {
 };
 
 const sendUpdateProgress = (tabId, payload) => {
-  //let command = payload[0];
     /*
      * let's invoke the experiment api in order to update the
      * model/engine download progress in the appropiate infobar
@@ -736,12 +708,4 @@ const sendUpdateProgress = (tabId, payload) => {
       tabId,
       localizedMessage
     );
-  /*browser.tabs.sendMessage(
-    tabId,
-    {
-      command,
-      tabId,
-      payload: [command, message]
-    }
-  );*/
 }
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -491,29 +491,29 @@ Promise.allSettled([displayedConsentPromise, isMochitestPromise]).then(values =>
   }
 });
 
-// return language models as urls for language pairs
+// return language models (as blobs) for language pairs
 const getLanguageModels = async (tabId, languagePairs) => {
   let start = performance.now();
 
-  let languageModelURLPromises = [];
+  let languageModelPromises = [];
   // eslint-disable-next-line no-use-before-define
-  languagePairs.forEach(languagePair => languageModelURLPromises.push(getLanguageModelAsURL(tabId, languagePair)));
-  let languageModelURLs = await Promise.all(languageModelURLPromises);
+  languagePairs.forEach(languagePair => languageModelPromises.push(getLanguageModel(tabId, languagePair)));
+  let languageModels = await Promise.all(languageModelPromises);
   let end = performance.now();
 
   console.log(`Total Download time for all language model files: ${(end - start) / 1000}s`);
   getTelemetry(tabId).record("timespan", "performance", "model_download_time_num", end-start);
 
   let result = [];
-  languageModelURLs.forEach((languageModelURL, index) => {
+  languageModels.forEach((languageModel, index) => {
     let clonedLanguagePair = { ...languagePairs[index] };
-    clonedLanguagePair.languageModelURL = languageModelURL;
+    clonedLanguagePair["languageModelBlobs"] = languageModel;
     result.push(clonedLanguagePair);
   });
   return result;
 };
 
-const getLanguageModelAsURL = async (tabId, languagePair) => {
+const getLanguageModel = async (tabId, languagePair) => {
   let languageModelPromise = [];
   languageModelFileTypes
       .filter(fileType => fileType !== "qualityModel" || languagePair.withQualityEstimation)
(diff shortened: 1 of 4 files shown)


lonnen:
I'm not one of the requests reviewers, but I grabbed the patch, built it, and confirmed this fixes the issue. Tested with Nightly 102.0a1 (2022-05-15) (64-bit).

abhi-agg:
@lonnen Thanks for confirming. I am not able to add you as a reviewer.

abhi-agg:
I am merging this one as I would like to request QA to confirm that this fix resolves the issue. Please feel free to file another issue if something doesn't look right here.

abhi-agg:
@andrenatal @Rob--W Would one of you be able to review this? I would like to merge it to request QA to confirm that this fix resolves the issue.

Rob--W (dismissed):
This is not the best way to solve the issue.

Translations in private browsing mode (and I guess also on http) are broken because some APIs are not available (`caches` in private browsing - #330, apparently `crypto` on http - #283).

This PR tries to fix the issue in private browsing issue by falling back to downloading the model directly. This will happen for every translation request in private browsing mode, which is very wasteful.

I suggest to move the downloading logic to the extension's background script, to have the following architecture:

1. `getItemFromCacheOrWeb`: worker posts message to the caller (who constructed `new Worker`).
2. Caller passes message to the background page (via `browser.runtime.sendMessage`)
3. Background page reads model, and validates checksum if needed. Note: if the URL can be assumed to never be re-used, it suffices to only validate the checksum after downloading it, and there is no need to validate it again upon reading it from the cache.
4. Background page returns response as a `Blob` to the caller (the content script from step 2).
5. Content script passes `Blob` to the Worker.
6. Worker uses the `FileReader` API to read the raw bytes behind it.

That would fix #330 and #283 at once.

Rob--W:
> @Rob--W Thanks for the comment. I was under the impression that the unavailability of `caches` API in private mode has nothing to do with where it is being used (a content/background script or a web worker).

This impression is incorrect. "Private browsing mode" refers to browser windows being in private browsing mode. 

> Is `caches` API available in the background script for private browsing?

Extensions with private browsing access are able to see activity from private browsing and regular browsing mode, and the background script would run in the context of regular browsing mode.

Firefox also has a "permanent private browsing mode" (opt in for users), in which regular browsing mode is disabled, and extensions (including the background page) would run in private browsing mode. In that case, the `caches` API won't be available either. To avoid fetching repeatedly in the fall back case, it may be useful to keep the data in the memory of the background page.

> Same question for `crypto`.

Yes.

abhi-agg:
> Firefox also has a "permanent private browsing mode" (opt in for users), in which regular browsing mode is disabled, and extensions (including the background page) would run in private browsing mode. In that case, the `caches` API won't be available either. To avoid fetching repeatedly in the fall back case, it may be useful to keep the data in the memory of the background page.

This is a useful information. It means we shouldn't be using `caches` API at all but implement our own simple cache (a `map` object to store everything in memory) if we want to support both "permanent private browsing mode" and "Private browsing mode". Right?

abhi-agg:
Documenting some of the discussions that happened b/w @Rob--W and me offline.

From performance point of view, the difference b/w the old and new architecture is following:
- The new architecture involves an additional step of passing the `Blob` of each model file from background to content script and then converting them to raw bytes (using FileReader API) in content script (Worker in our case)
- Cached files will be stored in memory in new architecture while old architecture uses browser cache that uses file system for the storage

I was wondering the overall impact of these differences in the following 2 scenarios:
1. When the files were downloaded from the N/w (e.g. the first time for any particular language pair)
2. When the files were already available in cache and didn’t require download from N/w (e.g. subsequent times for the same language pair)

What I understood from our discussion is following:
1. The new architecture has very negligible impact on performance in scenario 1 (because N/w downloads cause way more latency/performance penalty).
2. For scenario 2, passing around `Blob` object itself might be a bit expensive. Therefore, creating a URL from it and passing this URL string to the Worker would save memory copies of the actual data being passed around.

@Rob--W Please correct me if I am wrong here 👍🏾

abhi-agg:
Additional question.

https://github.com/mozilla/firefox-translations/issues/283 is happening because `crypto.subtle` API is only available for secure context and that's why it fails for http pages (as per https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle).

You mentioned that the new architecture will also resolve this. Does it mean that `crypto.subtle` will be available in background scripts for http pages?

Rob--W:
> Additional question.
> 
> https://github.com/mozilla/firefox-translations/issues/283 is happening because `crypto.subtle` API is only available for secure context and that's why it fails for http pages (as per https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle).
> 
> You mentioned that the new architecture will also resolve this. Does it mean that `crypto.subtle` will be available in background scripts for http pages?

There is no "background script for http pages".

The extension's background script is independent of the web page. There is only one background page instance of an extension, and content scripts can communicate with it.

As mentioned elsewhere (not sure whether in an issue/PR or in chat), the extension document (loaded from a moz-extension:-URL) is considered a secure context, and will therefore have access to the crypto API.

andrenatal:
@Rob--W I met with @abhi-agg today and we spoke about this.

We'll move the download/cache retrieval step to the background script and pass the model to the worker in the page as whatever is more appropriate, in order to support private browsing and http pages, and then file another issue with a lower priority that does not block the update to discuss the best way to support permanent private browsing and have it implemented later, since that will require more tooling to manage the occasional in-memory cache mechanism and we don't have much time and hands to do that. 

I also spoke to mixedpuppy some time ago about keeping large memory objects in the bgscript and heard that could degrade the performance quickly, since just two languages with pivoting+outboud translations could easily cost you +100mb,  so we'll need to discuss that implementation carefully.

abhi-agg:
I made following changes (as proposed above):
1. Moved the code to download and cache model files from `translationWorker.js` to `backgroundScript.js`
2. `backgroundScript.js` is responsible to create a URL from the `Blob` of downloaded/cached model buffer (using `URL.createObjectURL(new Blob([buffer]))` and pass it to `translationWorker.js`
3. `translationWorker.js` calls `fetch` API on the received URL to get the model bytes

This workflow runs successfully for normal browsing mode. But it fails for "private browsing mode" while executing `fetch` in step 3 with the exception: `TypeError: NetworkError when attempting to fetch resource.`

@Rob--W Do you have any idea why it might be happening?

abhi-agg:
I made following changes (as proposed above):
1. Moved the code to download and cache model files from `translationWorker.js` to `backgroundScript.js`
2. `backgroundScript.js` is responsible to create a URL from the `Blob` of downloaded/cached model buffer (using `URL.createObjectURL(new Blob([buffer]))` and pass it to `translationWorker.js`
3. `translationWorker.js` calls `fetch` API on the received URL to get the model bytes

This workflow runs successfully for normal browsing mode. But it fails for "private browsing mode" while executing `fetch` in step 3 with the exception: `TypeError: NetworkError when attempting to fetch resource.`

@Rob--W Do you have any idea why it might be happening?

andrenatal:
This is the actual issue that's being displayed in the browser console window:

`Security Error: Content at moz-extension://293da5b7-6787-4b83-93f0-b509d893a258/controller/translation/translationWorker.js may not load data from blob:moz-extension://293da5b7-6787-4b83-93f0-b509d893a258/7c550256-6fda-4daa-a9c6-29b4a9c68d9a.` 

So it seems some CSP or OOP condition is still preventing it from working on private browsing windows?

rpl:
> This is the actual issue that's being displayed in the browser console window:
> 
> `Security Error: Content at moz-extension://293da5b7-6787-4b83-93f0-b509d893a258/controller/translation/translationWorker.js may not load data from blob:moz-extension://293da5b7-6787-4b83-93f0-b509d893a258/7c550256-6fda-4daa-a9c6-29b4a9c68d9a.`
> 
> So it seems some CSP or OOP condition is still preventing it from working on private browsing windows?

@andrenatal I just took a look into why the SecurityError is being triggered, and it is being raised from [BasePrincipal::CheckMayLoadHelper](https://searchfox.org/mozilla-central/rev/9f95c41a962c9228f569f8a6b2c30edbb50b65ae/caps/BasePrincipal.cpp#610-612) with the following stack trace:
```
#16 0x00007fc11ad1d11c in mozilla::BasePrincipal::CheckMayLoadHelper(nsIURI*, bool, bool, unsigned long)
(this=0x7fc108f8cce0, aURI=0x7fc10f1fb350, aAllowIfInheritsPrincipal=false, aReport=true, aInnerWindowID=0) at /.../caps/BasePrincipal.cpp:610
#17 0x00007fc11ad296dd in nsScriptSecurityManager::CheckLoadURIWithPrincipal(nsIPrincipal*, nsIURI*, unsigned int, unsigned long)
(this=0x7fc1158037c0, aPrincipal=0x7fc108f8cce0, aTargetURI=<optimized out>, aFlags=0, aInnerWindowID=18446744073709551615)
at /.../caps/nsScriptSecurityManager.cpp:762
#18 0x00007fc11a8de84b in nsCORSListenerProxy::UpdateChannel(nsIChannel*, DataURIHandling, UpdateType)
(this=this@entry=0x7fc10f1f80d0, aChannel=aChannel@entry=0x7fc108f75468, aAllowDataURI=aAllowDataURI@entry=DataURIHandling::Allow, aUpdateType=aUpdateType@entry=UpdateType::Default) at /.../netwerk/protocol/http/nsCORSListenerProxy.cpp:926
#19 0x00007fc11a8de5e0 in nsCORSListenerProxy::Init(nsIChannel*, DataURIHandling) (this=0x7fc10f1f80d0, aChannel=0x7fc108f75468, aAllowDataURI=DataURIHandling::Allow)
at /.../netwerk/protocol/http/nsCORSListenerProxy.cpp:417
#20 0x00007fc11ca332d3 in DoCORSChecks(nsIChannel*, nsILoadInfo*, nsCOMPtr<nsIStreamListener>&) (aChannel=0x7fc1171f491d,
aChannel@entry=0x7fc108f75468, aLoadInfo=aLoadInfo@entry=0x7fc123f9a930, aInAndOutListener=...)
at /.../dom/security/nsContentSecurityManager.cpp:431
#21 0x00007fc11ca321de in nsContentSecurityManager::doContentSecurityCheck(nsIChannel*, nsCOMPtr<nsIStreamListener>&) (aChannel=0x7fc108f75468, aInAndOutListener=...)
at /.../dom/security/nsContentSecurityManager.cpp:1314
#22 0x00007fc11a47f467 in nsBaseChannel::AsyncOpen(nsIStreamListener*) (this=0x7fc108f75430, aListener=<optimized out>)
at /.../netwerk/base/nsBaseChannel.cpp:649
#23 0x00007fc11c28cbf8 in mozilla::dom::FetchDriver::HttpFetch(nsTSubstring<char> const&) (this=this@entry=0x7fc108e1cee0, aPreferredAlternativeDataType=...)
at /.../dom/fetch/FetchDriver.cpp:837
#24 0x00007fc11c28a22c in mozilla::dom::FetchDriver::Fetch(mozilla::dom::AbortSignalImpl*, mozilla::dom::FetchDriverObserver*)
(this=this@entry=0x7fc108e1cee0, aSignalImpl=aSignalImpl@entry=0x0, aObserver=<optimized out>) at /.../dom/fetch/FetchDriver.cpp:470
```

I managed to record the issue locally with rr and dig into it a bit more, and what I've seen seems to confirm my initial suspicious:

- the Blob URL seems to have been created from the extension background page, which (unless Firefox is running in Permanent Private Browsing mode) would have an extension principal with the privateBrowsing flag set to 0 in its originAttributes (to signal that the extension page isn't running in private browsing mode)

- the translationWorker.js that is fetching the Blobl URL (which I think [it is being triggered from here](https://github.com/mozilla-extensions/firefox-translations/blob/ef6ac0d2f127521c6b234519353769970b909296/extension/controller/translation/translationWorker.js#L58)) is running with privateBrowsing flag set to 1 (which I confirmed by [inspecting the value set on `mOriginAttributes.mPriviateBrowsingId` from the call to CheckMayLoadHelper](https://searchfox.org/mozilla-central/rev/9f95c41a962c9228f569f8a6b2c30edbb50b65ae/caps/BasePrincipal.cpp#612) in the session recorded with rr), and the fact that translationWorker.js is a web_accessible_url also suggest that the worker is being executed in the child private browsing page

Unfortunately in Firefox we don't support the "incognito split mode" (that in chrome creates two separate background pages, one in regular non-private browsing mode and the other one in private browsing mode) and so that isn't an option to deal with this use case.

Also, even if Firefox would be supporting the incognito split mode, there is also another way to trigger the same kind of issue (originAttributes mismatch between the extension principal that has created the Blob URL and the extension principal in the child web content process that is loading the BlobURL) in non private browsing tabs: **container tabs**.

I suspected that a container tab was also going to trigger the same issue, and so while I was here I gave that a try and verified that was the case as I was suspecting: with container tabs created with one of the extensions enabling the Firefox containers feature (e.g. Mozilla's MultiAccountContainers and Facebook Container, and some Tab Managers as well, like TreeStyleTabs), the exact same SecurityError is being raised on fetching the Blob url from the extension principal associated with the contentscript attached to the container tab (which is going to have in its originAttributes the same userContextId of the page that it is attached to, like the one attached to the private tab has the same privateBrowsing one associated to that page, while the background page's extension principal does have the default "non contained" userContextId).

And so the split more wouldn't help with that use case (as we couldn't have one more background page instance for every single container being created and/or in use).  

It would be worth making sure we have an issue on bugzilla to track a proper long term fix for this kind of use case (e.g. we could evaluate if relaxing that a bit for the extensions, allowing to load the Blob url associated to an extension principal even if the originAttributes mismatch at least if the extension has the private browsing access permission, is something that may be worth considering), I didn't find one still open that fits this one, would you mind to file one?

In the meantime, I was trying to think of any possible workaround on the extension side, and at the moment the only one I could think of is to try (maybe only for private and container tabs, where we know we are going to hit the SecurityError) to create the Blob and Blob URL on the content script side instead of creating it from the background page (which unfortunately would require to send that data to the content script over IPC using the WebExtensions messaging APIs).

abhi-agg:
@rpl Thanks a lot for debugging and sharing the root cause of this issue. This is super helpful and will save us lot of time 👍🏾 

> In the meantime, I was trying to think of any possible workaround on the extension side, and at the moment the only one I could think of is to try (maybe only for private and container tabs, where we know we are going to hit the SecurityError) to create the Blob and Blob URL on the content script side instead of creating it from the background page (which unfortunately would require to send that data to the content script over IPC using the WebExtensions messaging APIs).

I was experimenting a bit since morning. Instead of passing the Blob URL, I passed Blob from background script to the content script and read the Blob on the content script side. This worked without any problem.

However, as per my understanding, doing so will (at least theoretically) impact the overall performance as we are now passing huge amount of data (instead of a URL) via messaging APIs. Right?

(Btw, I don't see any noticeable difference while performing translation using this approach. But this is purely subjective.)

abhi-agg:
I tested this PR and it resolves both of the issues https://github.com/mozilla/firefox-translations/issues/324 and https://github.com/mozilla/firefox-translations/issues/283 👍🏾

rpl:
> @rpl Thanks a lot for debugging and sharing the root cause of this issue. This is super helpful and will save us lot of time 👍🏾
> 
> > In the meantime, I was trying to think of any possible workaround on the extension side, and at the moment the only one I could think of is to try (maybe only for private and container tabs, where we know we are going to hit the SecurityError) to create the Blob and Blob URL on the content script side instead of creating it from the background page (which unfortunately would require to send that data to the content script over IPC using the WebExtensions messaging APIs).
> 
> I was experimenting a bit since morning. Instead of passing the Blob URL, I passed Blob from background script to the content script and read the Blob on the content script side. This worked without any problem.
> 
> However, as per my understanding, doing so will (at least theoretically) impact the overall performance as we are now passing huge amount of data (instead of a URL) via messaging APIs. Right?
> 
> (Btw, I don't see any noticeable difference while performing translation using this approach. But this is purely subjective.)

@abhi-agg I had a chat with @bakulf on the topic, because I wasn't actually 100% convinced that passing a Blob over messaging would be sending its entire data via IPC right away.

Based on what I gathered from discussing about it with Andrea:
- IPCBlob would be trying to copy as less data as possible, and the data should be copied in chunks unless the Blob is < 1Mb
- WebExtensions messaging API will be serializing the engine messages into a StructuredCloneHolder and then deserialize them on the other side, that should be creating a new StreamBlob that points to the deserialize IPCBlobInputStream (or to a MemoryBlobInputStream if the entire message was smaller than 1Mb)

And so it seems that passing the Blob shouldn't force the entire data to go through IPC all at once neither.

I think that it may still be good to actually test that out (e.g. even just by creating a "reduced test case" that sends an artificially big blob over the WebExtensions messaging API and collect some profiles to confirm what is the actual impact in terms of memory usage), mainly to confirm that none of what the WebExtensions internals are doing under the hood for the messaging APIs are changing these assumption.

andrenatal:
I've just tested and finished reviewing this PR and, in addition to fixing the issues with private browsing and http, I noticed a significant performance improvement loading the model and the engine in a very constrained environment that I needed to set to this this (https://github.com/mozilla/firefox-translations/issues/336#issuecomment-1133168406). Also, the download progress in the infobar is now smoother and responsive. 

Thanks for helping out here at the last minute like always @rpl. I just spoke to Abhi and your explanation regarding the memory allocation corresponds to what we observed in our tests.

Thanks to everyone who helped in this thread specially @Rob--W for designing the best architecture to be adopted. 

This is going to be a great improvement to the product. Am merging it now.

andrenatal (approved):

PR #351, opened 2022-06-01
Bump version to 1.1.3

eu9ene:


lonnen (approved):

abhi-agg (approved):

eu9ene:
We should probably wait for other fixes before bumping the version and then release them all together

abhi-agg:
Yes. I was about to say the same. Specifically, I would like to include https://github.com/mozilla/firefox-translations/pull/353 as well.

PR #354, opened 2022-06-02
Add metadata to survey link

### LINKED ISSUE #352: Add extension version to the survey link
When we will make significant improvements in the extension, it will be important to distinguish responses in the survey by version.

https://help.alchemer.com/help/url-variables I think this should work if we just add `?version={extension_version}` to the survey URL.

eu9ene:
Added:
- extension version
- source language of last translation
- target language of last translation

This will help us to understand what version the feedback applies to and what language pairs a user translated last if there are complaints about translation quality.

fixes #352

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -11,6 +11,7 @@ modelRegistryRootURL, modelRegistryRootURLTest, modelRegistry */
  * extension's background process.
  */
 
+const extensionVersion = browser.runtime.getManifest().version;
 const scrubSentryEvent = ev => {
 
   /*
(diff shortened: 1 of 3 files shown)


abhi-agg:
@eu9ene Extension version information is definitely useful. However, I am not sure about the last source/target language used. A user might have run a series of translations on websites (possibly with different languages) before filing the survey. In this case, it will be misleading.

eu9ene:
> @eu9ene Extension version information is definitely useful. However, I am not sure about the last source/target language used. A user might have run a series of translations on websites (possibly with different languages) before filing the survey. In this case, it will be misleading.

Because of that I specifically state here and in reporting that this is the pair of the last translation. It's probably better than nothing.

abhi-agg (approved):

PR #395, opened 2022-06-10
Manifest v3

jelmervdl:
Should ultimately solve #10.


jelmervdl:
Sorry, I clicked the wrong button too quickly again 😭

PR #399, opened 2022-06-10
Restrict minimum version to what's compatible with the wasm engine and report errors compiling

### LINKED ISSUE #370: AMD CPU wasm validation error: unrecognized opcode: fd c
Testing version 1.1.2buildid20220523.151256 on an old AMD Phenom II X6 processor gives the following console error:

    Using optimized gemm (mozIntGemm) implementation
    Loading Translation Engine failure.
    failed to asynchronously prepare wasm: CompileError: wasm validation error: at offset 22339: unrecognized opcode: fd c bergamot-translator-worker.js:526:11

andrenatal:
@abhi-agg I found that this error happens not only on processors without SSE4.1, but also on Fx versions earlier than 90. 

fixes #370

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -7,13 +7,13 @@
   "applications": {
     "gecko": {
       "id": "firefox-translations-addon@mozilla.org",
-      "strict_min_version": "79.0"
+      "strict_min_version": "90.0"
     }
   },
   "browser_specific_settings": {
     "gecko": {
       "id": "firefox-translations-addon@mozilla.org",
-      "strict_min_version": "79.0"
+      "strict_min_version": "90.0"
     }
   },
   "options_ui": {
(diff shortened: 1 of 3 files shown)


extension/manifest.json
```diff
@@ -7,13 +7,13 @@
   "applications": {
     "gecko": {
       "id": "firefox-translations-addon@mozilla.org",
-      "strict_min_version": "79.0"
+      "strict_min_version": "90.0"
```
Rob--W, line 10: If you're bumping the min supported version to 90, then this snippet for Firefox 87- can be removed: https://github.com/mozilla/firefox-translations/blob/61295c474e0e68d653b7aa561bececb78ab48c9a/extension/controller/experiments/TranslationBar/api.js#L10-L15

andrenatal:
@abhi-agg ping r?

PR #408, opened 2022-06-16
Replace crypto uuid

### LINKED ISSUE #406: `self.crypto.randomUUID` is being used on PingSender.js and the mediator but it's not available on Fx < 95.0
```
Uncaught (in promise) TypeError: self.crypto.randomUUID is not a function
    _load moz-extension://8d8cdb81-d1ec-df4c-9f35-fa1333ac4a21/model/telemetry/PingSender.js:39
```

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility

eu9ene:
fixes #406

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -36,6 +36,7 @@
       "settings.js",
       "controller/translation/bergamotTranslatorVersion.js",
       "model/telemetry/schema.js",
+      "model/uuid.js",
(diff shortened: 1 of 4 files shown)


eu9ene:
The tests pass successfully locally.

andrenatal (approved):

PR #409, opened 2022-06-16
Add Ukrainian models

### LINKED ISSUE #166: Dump of Ukranian-centric models
A dump of models from Helsinki: https://github.com/Helsinki-NLP/UkrainianLT/blob/main/translateLocally-models.json

"Additional ones are on the way. The quality might be a bit questionable but it is hard for me to judge."

These have some languages where we don't have en support yet, need to revise the pivoting assumptions.

eu9ene:
The models were not quantized using browsermt marian (it produces so-called alpha models) and should be used with `gemm-precision: int8shiftAll`. See [browsermt/marian-dev](https://github.com/browsermt/marian-dev/blob/199201eb89b2941afdadb14164e936d412f897ad/src/common/config_parser.cpp#L927) config docs for an explanation of the difference. This mode is ~15% slower https://github.com/mozilla/translation-service/pull/20#issuecomment-1152880275. 

Also, we need to update Gecko to support non-alpha models, so they are slow for now ( I assume the same as usage on ARM).

We distinguish between different model types based on naming convention in the model registy.

The models include two different vocabularies for source and target language, unlike all the other models that currently reuse one vocabulary. I added two registry keys to support this.

Also, the model itself is a bit heavier and there's more data to download because of two vocabularies.

Despite all that it's good to have an extra language. It's marked as "Beta", so there shouldn't be expectations for speed or quality.

fixes #166

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -72,7 +72,7 @@ let translationRequestsByTab = new Map();
 let outboundRequestsByTab = new Map();
 const translateAsBrowseMap = new Map();
 let isMochitest = false;
-const languageModelFileTypes = ["model", "lex", "vocab", "qualityModel"];
+const languageModelFileTypes = ["model", "lex", "vocab", "qualityModel", "srcvocab", "trgvocab"];
 const CACHE_NAME = "fxtranslations";
 
 const init = () => {
@@ -509,6 +509,9 @@ const getLanguageModels = async (tabId, languagePairs) => {
   languageModels.forEach((languageModel, index) => {
     let clonedLanguagePair = { ...languagePairs[index] };
     clonedLanguagePair.languageModelBlobs = languageModel;
+    clonedLanguagePair.precision = modelRegistry[clonedLanguagePair.name].model.name.endsWith("intgemm8.bin")
+      ? "int8shiftAll"
+      : "int8shiftAlphaAll";
     result.push(clonedLanguagePair);
   });
   return result;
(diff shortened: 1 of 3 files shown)


andrenatal:
There's so many caveats that I don't know if we should implement it now.

andrenatal:
I will test it tomorrow morning.

eu9ene:
It looks pretty usable to me on a regular website like this one https://www.pravda.com.ua/

Of course, if you will try to translate huge texts on Wikipedia it will be noticeably slow but it's the same for our other models.

andrenatal:
I've pulled and tested the xpi built from CI and Ukrainian is not being displayed in the translation bar at least in a pt-br browser.

andrenatal (changes requested):
Translation bar does not display Ukrainian in a pt-br browser

andrenatal:
Same on english browser
Screen Shot 2022-06-23 at 4 20 04 PM

eu9ene:
> Translation bar does not display Ukrainian in a pt-br browser

did other languages work? I can see it working using `npm run once` but when I do `npm build` and then install it manually the translation bar does not appear at all

andrenatal:
> > Translation bar does not display Ukrainian in a pt-br browser
> 
> did other languages work? I can see it working using `npm run once` but when I do `npm build` and then install it manually the translation bar does not appear at all

yes
Screen Shot 2022-06-23 at 4 33 12 PM

eu9ene:
not sure what the issue was but I synced it with main and it looks working now

andrenatal (approved):

PR #420, opened 2022-06-28
Browse directly to Firefox Translations section on about:preferences when clicking on Options -> Translation Preferences

### LINKED ISSUE #319: Add a data-subcategory to make the preferences menu item work better
"Translation Preferences" on the Options buttons takes you to the generic about:preferences page, with the actual translation item buried in the middle. You can use the search box (if you know how that works) but it would be better if the menu item took you directly to the right place.

If you add a `data-subcategory="translations"` attribute to the added preference item you can then have the menu item target `about:preferences#general-translations`. If you put attribute on the hbox the whole line gets highlighted for a second or two, or you could try putting it on the button itself to highlight just that. I found the box more noticeable

andrenatal:
fixes #319 

along https://phabricator.services.mozilla.com/D150446


abhi-agg:
@andrenatal Just confirming. This change will work only if https://phabricator.services.mozilla.com/D150446 lands and rides the trains. Right?

abhi-agg (approved):

PR #432, opened 2022-07-06
Disables translation of forms if not supported by the model set

### LINKED ISSUE #384: Model downloading is stuck if translation of forms is enabled but reversed translation is not supported
**Steps**:
1. Go to https://nn.wikipedia.org/wiki/Norsk or any other page in Norwegian Nynorsk using English Firefox
2. Enable translation of forms
3. Press "Translate"

**Expected**:
- Checkbox "Enable translation of forms" is disabled for one-way languages
- Translation works

**Actual**:
- User is able to enable the checkbox and there is an error because it tries to download the reversed model that is required for the translation of forms but does not exist

**How it looks like:**

If using the language for the first time:

Screen Shot 2022-06-06 at 4 50 41 PM

If the model is already cached:

Screen Shot 2022-06-06 at 4 50 02 PM

andrenatal:
fixes #384

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-case-declarations */
 /* eslint-disable max-lines */
 /* global LanguageDetection, browser, PingSender, BERGAMOT_VERSION_FULL,
 Telemetry, loadFastText, FastText, Sentry, settings, deserializeError,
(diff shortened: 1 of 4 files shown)


extension/controller/backgroundScript.js
```diff
@@ -213,6 +214,14 @@ const messageListener = function(message, sender) {
           /*
            * request the experiments API do display the infobar
            */
+
+          let from = "en".concat(message.languageDetection.pageLanguage);
+          let to = message.languageDetection.navigatorLanguage.substring(0,2).concat("en");
+          if (from === "enen") from = to;
```
lonnen, line 220: a developer's comment explaining 220 and 221 would be helpful. It's not obvious why "enen" is a special case

lonnen (approved):
Suggest adding a comment. Otherwise this lgtm 👍

lonnen (approved):

Issue #467, opened 2022-07-26
Option to stop translation and show original text


Yona-TYT:
Currently I have problems posting on twitter when I activate the translator, because I needed to write in English but the plug-in translates everything I write into Spanish.

A button to stop the translation of the site momentarily while I write a post, that would be nice and very useful.

Greetings and thank you very much for this great plugin. :)

andrenatal:
Hi @Yona-TYT , you can switch off the option `Enable translation of forms`

Yona-TYT:
@andrenatal 
The "Enable translation of forms" option doesn't seem to be available on the twitter page for some reason.

Here on github if it appears

Yona-TYT:
[Grabación de pantalla desde 26-07-22 14:23:18.webm](https://user-images.githubusercontent.com/27737574/181083446-505422bc-1679-4d9f-9910-8389f5d8da13.webm)

I have also noticed that it is impossible to write on twitter while the translation is activated, see the attached video.

Yona-TYT:
I should also mention that typing is impossible, because the typing bar moves to the left constantly when translation is on.

andrenatal:
Yes, that's a real issue that I can reproduce. Problem is that the input field on twitter is actually a `span` element, which the parser thinks it's actually a translatable element.

jelmervdl:
For some reason I forgot to exclude all elements with contenteditable attributes? I've added exclusions for them in #476 which for me fixed Twitter. With that change you should be able to compose tweets without the browser fighting you.

Note that Twitter still looks a bit weird where not all tweets are always translated. That's because Twitter does its own language identification on tweets and adds `lang` attributes for each of them. If their language identification fails, and it did not match the source language the add-on expected, it will not translate the tweet. And for me their language identification fails quite often for shorter tweets, identifying English tweets as Romanian etc.

Yona-TYT:
@jelmervdl 
Thank you very much!, I will try it as soon as it is available.

By the way, currently every time I want to go back to the original text on any page I am forced to reload the page, is there a way to display the original text without reloading the page?, or maybe this hasn't been implemented yet?.

jelmervdl:
It hasn't been implemented yet.

I'm working on an initial implementation, but I'm expecting some trouble, especially with sites like Twitter that have dynamic content, or content changes that happened after the translation was inserted. For those pages, just reverting the state of the page to before the translation isn't going to be sufficient I expect.

andrenatal:
[...]

PR #476, opened 2022-07-28
Exclude `.notranslate` and `[contenteditable]` elements from translation

jelmervdl:
The `.notranslate` class is used by Google Translate to skip certain elements. And for obvious reasons we should not be changing the contents of elements with `contenteditable` enabled. This will fix the issue with composing tweets on Twitter mentioned in #467.


andrenatal:
ok, I tested this and it fixed but need to pass the linter. I'll do in a separate PR. Thanks @jelmervdl

jelmervdl:
Oops, sorry! I thought I caught all of the issues importing the fix, but I didn't actually run the linter 😅

PR #482, opened 2022-08-02
Revert "Remove obsolete key (applications) from manifest (#475)"

andrenatal:
This reverts commit 4f908f4a27b6e326f2d89035f46d9a074e98f9ff.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -4,6 +4,12 @@
   "description": "__MSG_extensionDescription__",
   "version": "1.1.4",
   "default_locale": "en_US",
+  "applications": {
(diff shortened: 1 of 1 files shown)


Rob--W:
FYI if the original goal of the patch was to avoid the log warning, then you can remove the `browser_specific_settings` key instead. The "applications" key predates `browser_specific_settings`, and as you have noticed, some tooling may work better with "applications" than "browser_specific_settings".

cadeyrn:
@andrenatal You should als remove the changelog entry for version 1.1.4 since you reverted the change.

PR #520, opened 2022-10-12
Bug 1794817 - Fix showing "translate as you browse" button.

emilio:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1794817.

display: inline didn't use to get honored before bug 1790920 (it ended up being equivalent to display: -moz-inline-box). Since the only thing we want is to remove the display: none in the style attribute, do that instead, letting the default display work.


andrenatal:
lgtm, but did not text.

PR #527, opened 2022-10-19
Stop relying on built-in .DTD and .properties for translations

flodolo:


extension/view/js/translation-notification-fxtranslations.js
```diff
@@ -252,17 +252,17 @@ window.MozTranslationNotification = class extends MozElements.Notification {
...
-    const bundle = Services.strings.createBundle("chrome://browser/locale/translation.properties",);
+
     let item = this._getAnonElt("neverForLanguage");
-    const kStrId = "translation.options.neverForLanguage";
+    let label = this.translationNotificationManager.localizedLabels.neverForLanguageLabel;
+
     item.setAttribute(
       "label",
-      bundle.formatStringFromName(`${kStrId}.label`, [langName]),
+      label.replace("%S", langName),
```
flodolo, line 261: This is flimsy, but as far as I understand this code doesn't have access to browser.i18n functions

andrenatal:
Thanks @flodolo, I can take it from here when I get back on Monday. If you could start landing the new strings in the l10n from Firefox would definitely help.

flodolo:
> If you could start landing the new strings in the l10n from Firefox would definitely help.

Done.

flodolo:
@andrenatal 
Ping? 

Note that the missing item in the menu is going to hit release users on Nov 15, and it's already hitting nightly and beta. Given the release process is not straightforward for this add-on, the sooner the fix lands, the better.

PR #543, opened 2022-11-02
Added option to translate to a target language different than the browser's

### LINKED ISSUE #357: Allow changing the language to translate to
it would be nice to allow translating to a language different from the one set in the browser. One use case is when you do not have the same native language as friends and want to translate a website in their language.

### LINKED ISSUE #430: Firefox Translations issues
Steps to reproduce:

    Open a webpage
    See translations bar above the page
    Check for source languages
    Check for target languages
    Check translation
    Try to see original page

Actual results:

    Cannot set source language same as Firefox language. For example you cannot set translation source language as German if your Firefox's language it set to German.
    Cannot set target languages. It always translates to English.
    Translations are bad. Also non-English characters are not being translated, but instead they were shown as weird characters.
    Once translation is completed, it is not possible to see original text

Expected results:

    I should be able to set source language, or addon should do it automatically.
    I should be able to set target language to whatever I want.
    Translations should be accurate and all characters should be in readable format.
    I should be able to revert page back to original state.

Original site
Translated site

### LINKED ISSUE #438: Allow selecting system language as translation source
Not being able to do that is bad for people who travel or for just checking the translation is correct (e.g. I'm bilingual but can't translate between the two languages i know because one is a system language) or people who like me just want to be able to arbitrarily select source and target language.

andrenatal:
fixes #357 
fixes #438
fixes #430

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ You can send suggestions or report issues to the team via any of the channels be
 - The display of the changelog now happens upon the user's choice on about:addons and in the options button.
 - Renewal of telemetry metrics
 - Normalization of language codes across the extension
+- Added option to translate to a target language different than the browser's
 
 ## [[v1.1.4]](https://github.com/mozilla/firefox-translations/tree/v1.1.4) - 2022-07-28
 ### Added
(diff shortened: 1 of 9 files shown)


sylvestre:
Terrific!

flodolo:
From a quick test, this would show up incorrectly in RTL, with the arrow going from target to source?

PR #557, opened 2022-11-08
Introduces option to prevent the translation bar from being displayed

### LINKED ISSUE #535: Option to turn off the banner
Hello,
it would be very nice if there was an option to permanently disable the banner under the address bar, as it is more often annoying than useful.

andrenatal:
fixes #535

### DIFF
--- a/extension/controller/backgroundScript.js
+++ b/extension/controller/backgroundScript.js
@@ -203,6 +203,10 @@ const messageListener = function(message, sender) {
             break;
         case "displayTranslationBar":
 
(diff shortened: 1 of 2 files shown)


andrenatal:
Depends on https://github.com/mozilla-l10n/firefox-translations-l10n/pull/17 to fully work.

flodolo:
This shouldn't go into the next release, you need to leave volunteer time to translate.

Unless I've missed one previous version already in review on AMO, since it needs to go out to users before 107 hits release (less than one week).

PR #630, opened 2023-01-13
Sanitize text

### LINKED ISSUE #573: Replace `innerHTML` by safer and more performant methods
https://github.com/mozilla/firefox-translations/blob/7c7f4fc667eee7b04b8e9a7ec85b2fb3e5501461/extension/view/popup/js/index.js#L82 and https://github.com/mozilla/firefox-translations/blob/7c7f4fc667eee7b04b8e9a7ec85b2fb3e5501461/extension/view/popup/js/index.js#L87 should be refactored into [safe DOM manipulation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Safely_inserting_external_content_into_a_page) that does not evaluate strings as code.

andrenatal:
fixes #573

### DIFF
--- a/extension/view/popup/index.html
+++ b/extension/view/popup/index.html
@@ -26,6 +26,7 @@
       </div>
       <div class="footer" id="status"></div>
     </div>
+    <script src="/model/static/DOMPurify/purify.min.js"></script>
     <script src="/controller/translation/TranslationMessage.js"></script>
     <script src="/model/Queue.js"></script>
     <script src="/model/contentErrors.js"></script>
--- a/extension/view/popup/index.html
+++ b/extension/view/popup/index.html
@@ -26,7 +26,6 @@
       </div>
       <div class="footer" id="status"></div>
     </div>
-    <script src="/model/static/DOMPurify/purify.min.js"></script>
     <script src="/controller/translation/TranslationMessage.js"></script>
     <script src="/model/Queue.js"></script>
     <script src="/model/contentErrors.js"></script>
--- a/extension/view/popup/js/index.js
+++ b/extension/view/popup/js/index.js
@@ -85,7 +85,7 @@ const setLangs = (selector, langsToSet, value, exclude) => {
         if (code !== exclude) {
             let name = langs.get(code);
             if (type === "dev") name += " (Beta)";
-            selector.innerHTML += DOMPurify.sanitize(`<option value="${code}">${name}</option>`, { USE_PROFILES: { html: true } });
+            selector.add(new Option(name, code));
         }
     }
     selector.value = value;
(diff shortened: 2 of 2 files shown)


extension/view/popup/js/index.js
```diff
@@ -79,12 +79,12 @@ class Mediator {
 }
 
 const setLangs = (selector, langsToSet, value, exclude) => {
-    selector.innerHTML = `<option value="0">${browser.i18n.getMessage("languageDefaultOption")}</option>`;
+    selector.innerHTML = DOMPurify.sanitize(`<option value="0">${browser.i18n.getMessage("languageDefaultOption")}</option>`, { USE_PROFILES: { html: true } });
```
Rob--W, line 82:
```suggestion
    selector.textContent = "";
    selector.add(new Option(browser.i18n.getMessage("languageDefaultOption"), "0"));
```

extension/view/popup/js/index.js
```diff
@@ -79,12 +79,12 @@ class Mediator {
...
     for (const [code, type] of Object.entries(langsToSet)) {
         if (code !== exclude) {
             let name = langs.get(code);
             if (type === "dev") name += " (Beta)";
-            selector.innerHTML += `<option value="${code}">${name}</option>`;
+            selector.innerHTML += DOMPurify.sanitize(`<option value="${code}">${name}</option>`, { USE_PROFILES: { html: true } });
```
Rob--W, line 87:
```suggestion
            selector.add(new Option(name, code));
```

extension/view/popup/index.html
```diff
@@ -26,6 +26,7 @@
       </div>
       <div class="footer" id="status"></div>
     </div>
+    <script src="/model/static/DOMPurify/purify.min.js"></script>
```
Rob--W, line 29:
```suggestion
```

Rob--W (changes requested):
The proposed use of DOMPurify is a bad fit. I suggest to use the `Option` constructor instead, to achieve exactly the same behavior in an efficient and readable way.

Relevant documentation:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/add

andrenatal:
> The proposed use of DOMPurify is a bad fit. I suggest to use the `Option` constructor instead, to achieve exactly the same behavior in an efficient and readable way.
> 
> Relevant documentation:
> 
>     * https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option
> 
>     * https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/add

@Rob--W ok maybe is better to then update this which was recommended in the PR https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Safely_inserting_external_content_into_a_page#html_sanitization

Rob--W (approved):

PR #639, opened 2023-01-15
Convert div to span due to inline semantics of span

veyndan:


andrenatal:
This one needs to be rebased

veyndan:
@andrenatal Rebased!

andrenatal (approved):

PR #648, opened 2023-01-24
Use the system-ui font-family for the popup

veyndan:
Makes the popup look a bit more at home on more platforms.

### DIFF
--- a/extension/view/popup/css/index.css
+++ b/extension/view/popup/css/index.css
@@ -7,7 +7,7 @@ body {
   height: 100%;
   margin: 0;
   font-size: 18px;
-  font-family: Optima, Helvetica, Arial;
+  font-family: system-ui;
 }
 
 body {


andrenatal (approved):

andrenatal:
Actually, I think I'll need to evaluate this one better. I'm afraid of breaking on some systems.

andrenatal:
I'll prefer to have a hardcoded font in this case, since we can't control whatever user's system's font can do to the UI here.

veyndan:
> I'll prefer to have a hardcoded font in this case, since we can't control whatever user's system's font can do to the UI here.

@andrenatal What would be an example of a user's system font changing the UI for the worse?

Just as an aside, but the current setup doesn't give us a whole lot of control over the user's fonts (unless we decide to start using remote fonts) as we're assuming that the client has one of `Optima, Helvetica, Arial` installed already.

PR #653, opened 2023-01-25
Create an android-win target in order to be able to load the extension on Windows

calixteman:


andrenatal (approved):
Did not test, but trust @calixteman that works!

package.json
```diff
@@ -10,11 +10,14 @@
     "build-test": "cp extension/settings/test.js extension/settings.js; npm run update-localization; web-ext build -s extension --overwrite-dest  --no-config-discovery -n firefox_translations.xpi",
     "lint:extension": "web-ext lint -s extension",
     "android": "cp scripts/manifest-dev-android.json extension/manifest.json; cp extension/settings/dev.js extension/settings.js; npm run update-localization; web-ext run --android-device=<device id from adb devices> -t firefox-android --firefox-apk org.mozilla.fenix -s extension --verbose --no-config-discovery --pref=extensions.experiments.enabled=true",
+    "android-win": "copy scripts\\manifest-dev-android.json extension\\manifest.json & copy extension\\settings\\dev.js extension\\settings.js & npm run update-localization-win & web-ext run --android-device=<device id from adb devices> -t firefox-android --firefox-apk org.mozilla.fenix -s extension --verbose --no-config-discovery --pref=extensions.experiments.enabled=true",
```
Rob--W, line 13: This cannot work. `<device id from adb devices>` needs to be edited manually... That does not belong to package.json

Rob--W (changes requested):

package.json
calixteman: I don't understand your point.
In the `android` target, there's the same `<device id from adb devices>` and the README says to manually modify it:
https://github.com/mozilla/firefox-translations#android

rpl: I believe that Rob want to pointing out that technically we could make the developer workflow slightly nicer by using a different approach to pass to web-ext the specific android device id that each developer will need, without having to manually edit the package.json (which may even make an actual device id to slip into a bigger commit because it may go unnoticed, given that the package.json is always going to be a revisioned file and changes to it added to a new commit unless explicitly excluded by the developer creating the commit).

As an example of other possible approaches, web-ext does also support passing values to set on cli arguments through environment variable (by convention the underlying library used to parse cli arguments, yargs, looks also for values to grab from environment variables, for this particular option if I'm not mistaken the computed name for the environment variables should be WEB_EXT_ANDROID_DEVICE), see this section of the doc for some more details about setting cli options through environment variables:
- https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#setting-option-environment-variables

There would be also another option, using a web-ext config file not committed in the repo, but for this particular use case the environment variables approach may be actually be simpler and nicer because there is no file to create (and an additional config file format to learn).

andrenatal: Thanks @rpl and @Rob--W . Would you please submit a PR with the correct approach then? Thanks!

rpl: > Thanks @rpl and @Rob--W . Would you please submit a PR with the correct approach then? Thanks!

@andrenatal sure thing, I gave it a try locally this morning and just opened https://github.com/mozilla/firefox-translations/pull/661

In the PR I'm proposing a slightly different approach, which uses the ability provided by `npm run` to pass additional options to the last command listed in the npm script (after a `--` to signal that the parameters that follows are not npm run options but options to pass down to the package.json script), that seems nicer than the environment variables given that the command to define it can change a bit based on the operating system and particular shell the developer may be used, while this approach should work across different the operating systems and command line shells.

btw, it is really great to be able to use it on Firefox for Android, I'm really looking forward to Firefox Translations be available to all Firefox users on android.

PR #663, opened 2023-01-28
Remove body wrapper

veyndan:
The `.app` is a redundant wrapper, as we can just style the `body` tag directly.

Tested on Android and Desktop.

### DIFF
--- a/extension/view/popup/index.html
+++ b/extension/view/popup/index.html
@@ -10,23 +10,21 @@
     />
   </head>
   <body>
-    <div class="app">
-      <div class="panel panel--from">
-        <label>
-          <select id="lang-from" name="from" class="lang-select"></select>
-        </label>
-        <textarea id="input" name="input" autofocus></textarea>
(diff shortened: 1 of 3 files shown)


extension/view/popup/index.html
```diff
@@ -10,23 +10,21 @@
...
-      <button class="swap" title="swap">↔️</button>
-      <div class="panel panel--to">
-        <label>
-          <select id="lang-to" name="to" class="lang-select"></select>
-        </label>
-        <textarea id="output" name="output" readonly></textarea>
-      </div>
-      <div class="footer" id="status"></div>
-    </div>
+		<div class="panel panel--from">
```
andrenatal, line 13: Thanks for the PR @veyndan . Can we just keep the indentation here?

andrenatal (changes requested):

extension/view/popup/index.html
veyndan: Done! Is that just to keep the git diff clean?

andrenatal: Thanks!

andrenatal (approved):

PR #734, opened 2023-07-06
Disable in-page translation functionality when the built-in version is enabled

marco-c:
The value of the pref is cached on purpose so we don't have to read it on every page load. The only downside is that users will have to restart Firefox for the pref change to take effect, but only few users twiddle with the prefs so it's really minor.


andrenatal (approved):
didn't test, but lgtm

PR #739, opened 2023-07-19
Manifest v3 compatibility

jelmervdl:
Get this thing to work both in Firefox and Chrome with manifest v3.

For chrome, we use chrome.offscreen to load a page that access the workers. Unfortunately, that means yet another layer of indirection.

Todo:
- [ ] restore state of tabs when service worker is unloaded/reloaded


jelmervdl:
Sorry, not my intention.

